PR #4, opened 2019-02-20
Feature/append list

GeorgFuechsle:
added a static function to append a List to any ListContainer at a specified position.
Including a test


derrg:
Is this story dependent on #3 being merged in? I see some identical changes in both PRs...Namely `Paragraph` and `ParagraphTest`.

svanteschubert:
If not someone else, I will take a look (doing a review) at all Simple API patches before the release, but as the Simple API contains quite a lot copied code from ODFDOM (as it was meant as a stand-alone fork), we are going to deprecate it and add usability functions - some kind of semantic API - to (or aside) of ODFDOM instead. So far the plan ;-)

PR #9, opened 2019-03-20
Use TravisCI too.

aadrian:


mistmist:
i'm not familiar with Travis CI but it sounds good to me.

There would be the option to use Jenkins CI on TDF infra (as is done for LibreOffice), but i guess that is heavily integrated with the TDF gerrit and not so much with GitHub, so probably it would need some extra work to set up (scripting & TDF admin) and Travis is going to be easier.

Are there any obvious downsides to Travis?

j-dimension:
TravisCI is restricted in terms of build runtimes. It might not be suitable for long running tests.
Also, when a build is run it spins up a VM in AWS, executes the build, destroys the VM. Logs are saved and viewable, but any build artefacts are gone. If you need them, the build needs to transfer it to some other server. There is no Jenkins workspace that you can browse after the build has finished.

It's a great tool with low maintenance though :-)

svanteschubert:
No objections from my side, if something doesn't work, we will adjust ;-)
Keep going! Thanks, Svante

Am Di., 16. Apr. 2019 um 16:48 Uhr schrieb mistmist <
notifications@github.com>:

mistmist:
status update: Travis CI is apparently enabled now - this needed approval from the "tdf" account; build fails currently but Svante is investigating it

PR #14, opened 2019-05-14
Maven cleanup

svanteschubert:
I have removed (hopefully) the final namings of Apache (aside the license) and updated the Maven pom.xml to a recent edition so the current [JDK 11 / 1.0.0-SNAPSHOT branch](https://github.com/svanteschubert/odftoolkit/tree/odf-changes) is using the same plugins as our JDK 8 master branch.

Our new mailing list is now available and added:
`<mailingList>`
`<!--`
`Help: <dev+help@odftoolkit.org>`
`Digest subscription: <dev+subscribe-digest@odftoolkit.org>`
`No-mail subscription: <dev+subscribe-nomail@odftoolkit.org>`
`-->`
`<post>dev@odftoolkit.org</post>`
`<subscribe>dev+subscribe@odftoolkit.org</subscribe>`
`<unsubscribe>dev+unsubscribe@odftoolkit.org</unsubscribe>`
`<archive>https://listarchives.odftoolkit.org/dev/</archive>`
`<name>dev.odftoolkit.org</name><!-- List-Id -->`
`</mailingList>`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,9 @@ local.properties
 .loadpath
 .recommenders
 
+### Netbeans ###
+nb-configuration.xml
+
 # External tool builders
 .externalToolBuilders/
 
@@ -195,3 +198,5 @@ buildNumber.properties
 .mvn/wrapper/maven-wrapper.jar
 
 # End of https://www.gitignore.io/api/java,maven,eclipse,intellij+all
+/xslt-runner/nbproject/private/
+/xslt-runner-task/nbproject/private/
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,5 +1,10 @@
-Apache ODF Toolkit Change Log
+ODF Toolkit Change Log
 =============================
+======= Release 0.9.0 =======
+
+First release at the Document Foundation, resembling the latest state of the sources from Apache ODF Toolkit.
+Release is based on JDK 1.8, the next release will be start working based on JDK 11.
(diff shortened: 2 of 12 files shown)


HEADER.txt
```diff
@@ -1,5 +1,5 @@
 /*
- * Licensed to the Apache Software Foundation (ASF) under one or more
+ * Licensed to the Document Foundation (TDF) under one or more
```
mistmist: the CLAs are still between ASF and contributors... please revert this

LICENSE
```diff
@@ -206,13 +206,13 @@
 
 ====================================================================================================
 
-ODF TOOLKIT DEPENDENCIES: 
+TDF ODF TOOLKIT DEPENDENCIES:
```
mistmist: hmm i don't see much value in pefixing TDF here ... at the ASF that was some sort of policy to prefix Apache to everything but it's not called "TDF LibreOffice" so why do it here?

NOTICE
```diff
@@ -1,18 +1,19 @@
 The ODF Toolkit
 
-Copyright 2011-2018 The Apache Software Foundation.
```
mistmist: i think we should not change any of this file

svanteschubert:
Thanks for your review, Michael. I have fixed the problems you mentioned and dispatched the whitespace handling problem into a separate issue: https://github.com/tdf/odftoolkit/pull/15

PR #19, opened 2019-07-20
Implementing JDK 7 autocloseable feature for OdfPackage and document classes

svanteschubert:
I found a cosmetic change that might be still become quite helpful for our upcoming release:
The implementation of the JDK 7 autocloseable feature, see
https://examples.javacodegeeks.com/core-java/java-autocloseable-interface-example/
https://docs.oracle.com/javase/8/docs/api/java/lang/AutoCloseable.html

Please review, if I have overseen anything important :-)

Thanks in advance,
Svante


svanteschubert:
Here is another site with information: https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

PR #20, opened 2019-07-31
Changes for writer table

uweTTC:
Hi
We try to use this lib, but we must make some changes. Our software generates some writer tables automatical and replaces the placeholder in the writer document with these tables. To format the tables, I make some changes on the lib.


svanteschubert:
Hi and thanks for the patch!

The builds are failing due to some whitespace problem, which can be fixed by calling 'mvn editorconfig:format' as noted in the end of the build log. 
I have tested it on the pull-request branch: https://github.com/tdf/odftoolkit/compare/pr/20
(you may checkout pull request as described in https://gist.github.com/piscisaureus/3342247)
But it does not seem to reach your pull request.. 
Anyway, no action required, I will merge the request and fix it afterwards..

Please note that the Simple API was once a fork by IBM from ODFDOM before rejoining the ODF Toolkit again. Unfortunately, the fork was never merged again by IBM and contains a lot of duplicated code. Therefore, we decided to deprecate it and remove it after the next release from the repository.

PR #32, opened 2020-04-24
Fix handling of column groups in Tables

kuhmuh:
Hello everyone,

I know the simple API is deprecated, but I'd like to fix some issues with groups inside.

Maybe your are interested.

Best regards


svanteschubert:
Hi thanks for the patch, I will review it! Anyone else interested in Simple API is asked to do the same and take a look at it!
As we have not yet released the final ODF Toolkit release 0.9.0 (the last Simple API) release it is fine.

In addition, I come to the conclusion that I would like to overtake most/some of the Simple API to ODFDOM. But not copying the copied part ;-) I will write more about - likely end of May - before (or shortly) I start(ed) evaluating it.

PR #37, opened 2020-04-28
Whitespace changes during build

smehrbrodt:
Somehow these files are changed during build, commit those changes so that we don't have a dirty state after each build.


svanteschubert:
Yes, this is a whitespace normalization so we will commit all same whitespace to the repository and have no 'syntax noise' when looking at DIFFs.
Incidentally, I have explained this yesterday here:
https://github.com/openpreserve/jhove/issues/516#issuecomment-619873078
(sorry for that bad layout, markdown was surprising me, I should have checked the preview before sending it)

Issue #26, opened 2020-01-20
Adding ODF 1.3 support to the ODF Toolkit


svanteschubert:
I have started a [branch of the 0.9.0 version to support ODF 1.3](https://github.com/tdf/odftoolkit/tree/0.9-ODF1.3).
but decided in discussions with Michael Stahl and Oliver Rau afterwards that it might be easier to have an ODF 1.3 support based solely on the upcoming ODF 1.0.0 version.
The reason for this is that 1.0.0 is the merge of a several-year-old merge form Open-XChange using ODFDOM for their backend of their new [OX Documents](https://www.open-xchange.com/portfolio/ox-documents/).

In the following, I am starting a listing what needs to be fixed - this is likely being updated during the lifetime of this issue.
This documentation should be added to our SITE documentation to describe how to do an ODF version change in the future.

Changes expected so far:

1. The generator project needs to generate new ODFDOM typed DOM classes from the new ODF XML schema/grammar files. These files need to replace the ODFDOM prior classes for elements and attributes.

2. The generator tests have to be adopted to embrace ODF 1.3.

3.  The ODF Validator needs to have an ODF 1.3 instead of an ODF 1.2 option.

4. The ODF templates within ODFDOM - being used when creating new documents from the scratch needs to be added to ODF 1.3. 
I did this usually by editing the XML within the ZIP using JEdit with its archive and its XML extension.

svanteschubert:
For simplicity reason, we are not providing yet the support of multiple versions at the time or the transformation of one version to the other.
IMHO this functionality should be described at the specification level, explaining the differences to the prior ODF version in a way that transformations can be generated.

svanteschubert:
5. The HTML form of the ODF specification has to be exchanged with all its images. 
Test the HTML specification - should already be done on TC level, but better be safe than sorry:
5a: A broken link check should be done 
5b: OLE objects in the specification should have been exchanged with images, otherwise they won't be visible in the HTML version - for instance by taking a look at into the ODT if there are still OLE objects.  
6. The taglet module, which is doing insertions of in the ODFDOM JavaDoc to the HTML form of the ODF specification has to be updated to reference to a new specification.

svanteschubert:
NOTE: The current ODF 1.3 branch based on the master will be abandoned and new ODF 1.3 branch will be created for the 1.0.0 release branch.

svanteschubert:
There are **36 new feature/enhancement in ODF 1.3**.
[...]

PR #40, opened 2020-04-30
ODF 1.3 new mimetype: Text master template

smehrbrodt:
Add validator support for ODF1.3 mimetype "application/vnd.oasis.opendocument.text-master-template" (OFFICE-2580)

### DIFF
--- a/odfdom/src/main/java/org/odftoolkit/odfdom/doc/OdfDocument.java
+++ b/odfdom/src/main/java/org/odftoolkit/odfdom/doc/OdfDocument.java
@@ -122,6 +122,7 @@ public abstract class OdfDocument extends OdfSchemaDocument {
 		TEXT("application/vnd.oasis.opendocument.text", "odt"),
 		TEXT_MASTER("application/vnd.oasis.opendocument.text-master", "odm"),
 		TEXT_TEMPLATE("application/vnd.oasis.opendocument.text-template", "ott"),
+		TEXT_MASTER_TEMPLATE("application/vnd.oasis.opendocument.text-master-template", "otm"),
 		TEXT_WEB("application/vnd.oasis.opendocument.text-web", "oth");
 		private final String mMediaType;
 		private final String mSuffix;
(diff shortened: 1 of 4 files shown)


svanteschubert:
I am afraid this pull request is dependent on the following enhancement:
https://github.com/tdf/odftoolkit/issues/26

The complete ODFDOM is using currently solely ODF 1.2.
Like the templates that will be used, when creating an empty document are hardcoded to ODF 1.2:
https://github.com/tdf/odftoolkit/tree/1.0.0_SNAPSHOT/odfdom/src/main/resources

You can not mix ODF 1.2 with ODF 1.3 functionality unless we have transform functionality between the versions.

mistmist:
oh... would it be possible to merge only the changes to validator and leave the odfdom as it is, or would that not work?

svanteschubert:
Sure, validator only should work for me! :-)

smehrbrodt:
> oh... would it be possible to merge only the changes to validator and leave the odfdom as it is, or would that not work?

Sadly that doesn't work. Produces the following error (which comes from odfdom, not validator):

`odf13_new_mimetype.odt:  Error: The ODF mimetype 'application/vnd.oasis.opendocument.text-master-template' is invalid for the ODF XML Schema document!`

smehrbrodt:
> > oh... would it be possible to merge only the changes to validator and leave the odfdom as it is, or would that not work?
> 
> Sadly that doesn't work. Produces the following error (which comes from odfdom, not validator):
> 
> `odf13_new_mimetype.odt: Error: The ODF mimetype 'application/vnd.oasis.opendocument.text-master-template' is invalid for the ODF XML Schema document!`

So what I could do if we don't want to change odfdom, is convert this Error into a warning. That way, validation would not fail, and we can have the validator supporting 1.3 while odfdom is still on 1.2

Would that work for you?

svanteschubert:
As discussed by phone a branch would be the best solution, as changing the error to warning would be a bug (every 'shall/must' of the ODF 1.2 specification not used is an error, every 'should' not used a warning). And adding the new mimetype in the existing ODF 1.2 ODFDOM would make it possible to create invalid ODF 1.2 with the upcoming release.

smehrbrodt:
Closing this in favor of #43.

PR #47, opened 2020-05-14
validator: add fallback validation of manifest.xml to work around Ope…

mistmist:
…nJDK bug

The problem is that ZipInputStream can't open the encrypted documents
that LO writes, you get an exception:
ZIP entry 'mimetype': only DEFLATED entries can have EXT descriptor

https://bugs.openjdk.java.net/browse/JDK-8143613

It's not possible to validate most of an encrypted document anyway, but it is
possible and very useful to validate the manifest.xml itself, because
(obviously) this is the only situation when it contains encryption related
markup.

The OdfPackage can't be used for this since it depends on the buggy
java.util.zip stuff.

So implement a fallback, when the expected ZipException arrives
OdfRootPackageValidator goes on to parse the ZIP central directory itself and
then reads META-INF/manifest.xml and validates only that.


mistmist:
wanted to add a test and noticed that the test reads not from File but from InputStream... guess i'll change it to read the File or Stream into byte[] and work on that, should give a generic impl and files are typically not large enough to cause trouble

PR #48, opened 2020-05-15
validator: add fallback validation of manifest.xml to work around Ope…

mistmist:
…nJDK bug

The problem is that ZipInputStream can't open the encrypted documents
that LO writes, you get an exception:
ZIP entry 'mimetype': only DEFLATED entries can have EXT descriptor

https://bugs.openjdk.java.net/browse/JDK-8143613

It's not possible to validate most of an encrypted document anyway, but it is
possible and very useful to validate the manifest.xml itself, because
(obviously) this is the only situation when it contains encryption related
markup.

The OdfPackage can't be used for this since it depends on the buggy
java.util.zip stuff.

So implement a fallback, when the expected ZipException arrives
OdfRootPackageValidator goes on to parse the ZIP central directory itself and
then reads META-INF/manifest.xml and validates only that.

Adapt ODFStreamValidator so that it copies the InputStream into a buffer
first; if OdfPackage.loadPackage is successful the buffer is discarded.

Add testcase validateEncryptedODF13WithInvalidOpenPGP.


svanteschubert:
Perhaps someone might give it a try (or at least evaluate) to fix the problem directly at the ODFDOM using a different ZIP implementation, like https://commons.apache.org/proper/commons-compress/zip.html

Please wait with the merge a week...

mistmist:
okay, this branch is not urgent, just wanted it in 0.9 so i can use it in LO test...

validator/src/test/java/org/odftoolkit/odfvalidator/InvalidPackageTest.java
```diff
@@ -124,7 +124,30 @@ public void validateEncryptedODT() {
...
+        Assert.assertTrue(output.contains("Fatal: only DEFLATED entries can have EXT descriptor"));
+        Assert.assertTrue(output.contains("Fatal: The document is encrypted. Validation of encrypted documents is not supported."));
+        java.util.logging.Logger.getLogger(getClass().getName()).log(Level.INFO, "Test result:\n{0}", output);
+    }
+
+    @Test
+    public void validateEncryptedODF13WithInvalidOpenPGP() {
+        String output = "";
+        try {
+            // password: hello
```
sebkur, line 136: Can't open the file using this password

sebkur:
Looking into this right now. @mistmist out of curiosity, what's the password for validator/src/test/resources/gpg13.odt ?

validator/src/test/java/org/odftoolkit/odfvalidator/InvalidPackageTest.java
mistmist: errr oops you're right thats copypasta from the previous test :(
there is no password, it requires a GPG key, i'm using this command to run LO and create the file, it will open without requiring a password; there's also some bug somewhere that causes LO to ask for a password when opening the file if it can't find the GPG key.

GNUPGHOME=/lo/build_icecc/workdir/CppunitTest/xmlsecurity_signing.test.user instdir/program/soffice

mistmist: but it's not really useful for this test to be able to open the document, adding GPG decryption to Validator seems definitely out of scope to me, it's enough if we can validate the manifest.xml which is never encrypted

sebkur: aha, yes, the password prompt from LibreOffice made me think it was encrypted with a password

svanteschubert:
The conflicts are likely caused by the code formatting.
Going to add the formatting plugin to the pom.xml of the branch and format them locally:
`mvn -X git-code-format:format-code -Dgcf.globPattern=**/*`

@sebkur You seem to have reviewed this pull request.
Is this fix otherwise fine so it can be merged?

sebkur:
@svanteschubert yes I reviewed it, however I could not yet figure it out. My work in progress is here: https://github.com/sebkur/odftoolkit/commits/change-zip-impl

Changing the ZIP implementation causes a lot of test cases to behave differently and it is unclear for each individual case whether that is actually an improvement or if it is now causing a bug.

sebkur:
somebody who is familiar with the test cases will need to look into this

sebkur:
@mistmist could you have a look at #53? Thanks

svanteschubert:
@mistmist We might want to consider exchanging the broken ZIP library instead of providing a workaround in our library.
Could you please review the approach @sebkur has suggested in #53 and tell us the pros&cons from your perspective?

sebkur:
> I'd be in favour of merging this - with #53 apparently taking more time (or not an option in the end)

I think this solution is rather hackish. @mistmist pointed out that this is not urgent so I'd suggest we wait and see if we can come to a non-hackish, solid solution using a well-working, standards-conforming ZIP-library as proposed originally in #53 and now continued in #58.

sebkur:
> …nJDK bug
> 
> The problem is that ZipInputStream can't open the encrypted documents
> that LO writes, you get an exception:
> ZIP entry 'mimetype': only DEFLATED entries can have EXT descriptor
> 
> https://bugs.openjdk.java.net/browse/JDK-8143613
> 
> It's not possible to validate most of an encrypted document anyway, but it is
> possible and very useful to validate the manifest.xml itself, because
> (obviously) this is the only situation when it contains encryption related
> markup.
> 
> The OdfPackage can't be used for this since it depends on the buggy
> java.util.zip stuff.
> 
> So implement a fallback, when the expected ZipException arrives
> OdfRootPackageValidator goes on to parse the ZIP central directory itself and
> then reads META-INF/manifest.xml and validates only that.
> 
> Adapt ODFStreamValidator so that it copies the InputStream into a buffer
> first; if OdfPackage.loadPackage is successful the buffer is discarded.
> 
> Add testcase validateEncryptedODF13WithInvalidOpenPGP.

I just got feedback on the [Commons Compress issue](https://issues.apache.org/jira/projects/COMPRESS/issues/COMPRESS-549) I created (https://github.com/tdf/odftoolkit/pull/53#issuecomment-659285991). It turns out the test case is probably bogus itself :D

mistmist:
oops i forgot to close this, thanks @sebkur for coming up with a better solution!

PR #53, opened 2020-06-28
Change ZIP implementation

sebkur:
In order to fix the problems described in #48 let's try to use a solid library for working with ZIP files, such as commons-compress. This is a work in progress PR, I'm looking for help. The unit tests are still failing and somebody familiar with the test cases will need to review what's going on there. Maybe the test cases are wrong (they describe the behavior we're actually trying to fix here?). Or something else is happening, I'm not sure at this point.


svanteschubert:
@sebkur Cool, many thanks for your contribution. 
@mistmist Could you take a look into it, Michael?

validator/src/test/java/org/odftoolkit/odfvalidator/InvalidPackageTest.java
```diff
@@ -32,16 +32,22 @@
...
+        } catch (Throwable t) {
+            t.printStackTrace();
+            Assert.fail(t.toString());
+        }
+        System.out.println(output);
+        Assert.assertTrue(output.contains("testInvalidPkg1.odt/mimetype:  Error: The file 'mimetype' is not the first file in the ODF package"));
+        Assert.assertTrue(output.contains("testInvalidPkg1.odt/mimetype:  Error: The file 'mimetype' shall not be compressed"));
+        Assert.assertTrue(output.contains("testInvalidPkg1.odt/mimetype:  Error: There shall be no extra field for the 'mimetype' file"));
+        Assert.assertTrue(output.contains("testInvalidPkg1.odt/META-INF/manifest.xml:  Error: The file 'Configurations2/accelerator/current.xml' shall not be listed in the 'META-INF/manifest.xml' file as it does not exist in the ODF package"));
+        Assert.assertTrue(output.contains("testInvalidPkg1.odt:  Info: 8 errors, 10 warnings"));
```
mistmist, line 50: i dont get 8 errors here?
both in the 0.9 branch "master" and with this branch i get "7 errors".

validator/src/test/java/org/odftoolkit/odfvalidator/InvalidPackageTest.java
```diff
@@ -145,44 +145,55 @@ public void validatePackage3() {
...
-            // password: hello
-            String name = "gpg13.odt";
-            output = doValidation(name, null);
-        java.util.logging.Logger.getLogger(getClass().getName()).log(Level.INFO, "Test result:\n{0}", output);
-        } catch (Exception t) {
-            StringWriter errors = new StringWriter();
-            t.printStackTrace(new PrintWriter(errors));
-            Assert.fail(t.toString() + "\n" + errors.toString());
-        }
-        Assert.assertTrue(output.contains("Fatal: only DEFLATED entries can have EXT descriptor"));
```
mistmist: this error indicates that the ZIP library has the bug referenced in https://github.com/tdf/odftoolkit/pull/48 i.e. it does not implement the current version of the ZIP specification.

switching to a different ZIP library only makes sense if it results in this error no longer happening.

validator/src/test/java/org/odftoolkit/odfvalidator/InvalidPackageTest.java
```diff
@@ -145,44 +145,55 @@ public void validatePackage3() {
...
-        Assert.assertTrue(output.contains("Fatal: The document is encrypted. Validation of encrypted documents is not supported."));
```
mistmist: this is an expected error message (we need to inform the user that we can't do what they asked for) that is currently a consequence of the previous error (iirc) - if the previous error is fixed it still needs to be produced but probably in a different way.

validator/src/test/java/org/odftoolkit/odfvalidator/InvalidPackageTest.java
```diff
@@ -145,44 +145,55 @@ public void validatePackage3() {
...
-        Assert.assertTrue(output.contains("META-INF/manifest.xml:  Info: ODF version of manifest: \"1.3\""));
```
mistmist: informational message, should remain.

validator/src/test/java/org/odftoolkit/odfvalidator/InvalidPackageTest.java
```diff
@@ -145,44 +145,55 @@ public void validatePackage3() {
...
-        Assert.assertTrue(output.contains("META-INF/manifest.xml[3,20]:  Error: tag name \"manifest:keyinfo\" is not allowed. Possible tag names are: <encrypted-key>,<file-entry>"));
```
mistmist: this error is produced by validating the manifest.xml file, which is the only file in an encrypted ODF package that is not encrypted - it must remain, that's the main point of this test.

validator/src/test/java/org/odftoolkit/odfvalidator/InvalidPackageTest.java
```diff
@@ -145,44 +145,55 @@ public void validatePackage3() {
...
-        Assert.assertTrue(output.contains("META-INF/manifest.xml[10,25]:  Error: element \"manifest:CipherData\" was found where no element may occur"));
```
mistmist: same as previous.

validator/src/test/java/org/odftoolkit/odfvalidator/InvalidPackageTest.java
```diff
@@ -145,44 +145,55 @@ public void validatePackage3() {
...
-        Assert.assertTrue(output.contains("META-INF/manifest.xml:  Info: 2 errors, no warnings"));
```
mistmist: summary of the previous 2 lines, so must also remain.

sebkur:
@mistmist thank you for the detailed explanations!

sebkur:
I tried to get to the bottom of this and see whether there is any Zip implementation for Java that can work with these files.

I set up a testing repository that focuses just on reading data from your example file here: https://github.com/sebkur/test-zip-impls

I created testing classes that try to read the entries of the file using the following classes:
* [java.util.zip.ZipInputStream from JDK11](https://github.com/sebkur/test-zip-impls/blob/master/src/test/java/de/topobyte/zip/tests/TestZipInputStream.java)
* [java.util.zip.ZipFile from JDK11](https://github.com/sebkur/test-zip-impls/blob/master/src/test/java/de/topobyte/zip/tests/TestZipFile.java)
* [org.apache.commons.compress.archivers.zip.ZipArchiveInputStream from `org.apache.commons:commons-compress:1.20`](https://github.com/sebkur/test-zip-impls/blob/master/src/test/java/de/topobyte/zip/tests/TestCommonsZipInputStream.java)
* [net.lingala.zip4j.io.inputstream.ZipInputStream from `net.lingala.zip4j:zip4j:2.6.1`](https://github.com/sebkur/test-zip-impls/blob/master/src/test/java/de/topobyte/zip/tests/TestZip4jInputStream.java)

# Results

## java.util.zip.ZipInputStream

Fails with a ZipException as expected.

## java.util.zip.ZipFile

Can read all entry metadata and extract the two unencrypted files `mimetype` and `META-INF/manifest.xml`.

## org.apache.commons.compress.archivers.zip.ZipArchiveInputStream

By default it also fails with `UnsupportedZipFeatureException`.

However, we can pass true as a parameter called `allowStoredEntriesWithDataDescriptor`. It behaves differently then.
It can parse the filenames contained in the archive and also succeeds in reading the file size and content for `mimetype`, however it fails with all subsequent file sizes and contents, here's the output of the test:

    entry: 'mimetype' compressed: 39, uncompressed: 39
    entry: 'styles.xml' compressed: -1, uncompressed: -1
    entry: 'content.xml' compressed: -1, uncompressed: -1
    entry: 'manifest.rdf' compressed: -1, uncompressed: -1
    entry: 'Configurations2/toolpanel/' compressed: 0, uncompressed: 0
    entry: 'Configurations2/menubar/' compressed: 0, uncompressed: 0
    entry: 'Configurations2/floater/' compressed: 0, uncompressed: 0
    entry: 'Configurations2/popupmenu/' compressed: 0, uncompressed: 0
    entry: 'Configurations2/images/Bitmaps/' compressed: 0, uncompressed: 0
    entry: 'Configurations2/accelerator/' compressed: 0, uncompressed: 0
    entry: 'Configurations2/statusbar/' compressed: 0, uncompressed: 0
    entry: 'Configurations2/toolbar/' compressed: 0, uncompressed: 0
    entry: 'Configurations2/progressbar/' compressed: 0, uncompressed: 0
    entry: 'meta.xml' compressed: -1, uncompressed: -1
    entry: 'settings.xml' compressed: -1, uncompressed: -1
    entry: 'META-INF/manifest.xml' compressed: -1, uncompressed: -1

## net.lingala.zip4j.io.inputstream.ZipInputStream

Doesn't throw any exceptions, just stops after the second entry reporting zero size:

    entry: 'mimetype' compressed: 39, uncompressed: 39
    entry: 'styles.xml' compressed: 0, uncompressed: 0

sebkur:
So it's pretty interesting, ZipFile seems to do everything right while ZipInputStream is broken. Also all zip libraries I could find out there seem to be broken too. Seems crazy. I can see that it is extremely hard to repair something in the official JDK code. And even if we were to get to repair that, we would have to wait ages for the changes to arrive in actual deployments of JDK and could not really rely on the fix being ther. However fixing commons-compress or zip4j seems a lot more realistic and by depending on one of those libraries with a version number that includes a fix to the problem would pull in a proper implementation into the odftoolkit.

sebkur:
I mean we can also just use your workaround from #48, however if the ODF files that LibreOffice produces are really conforming to the latest version of the ZIP implementation, and all the Zip libraries for Java out there are not working properly with such conforming ZIP files, I personally feel a strong urge to fix those libraries instead of going with a quick fix that solves this just for us. @mistmist Do you have any pointers to the relevant changes to the zip specification that introduced this feature that LibreOffice is using and the libraries and JDK do not support?

mistmist:
https://bugs.openjdk.java.net/browse/JDK-8143613 claims that the current APPNOTE.TXT says the limitation is from PKZIP version 2.04g for DOS, which i remember using in the mid 1990s.

there is an archive of APPNOTE.TXT here:
https://support.pkware.com/home/pkzip/developer-tools/appnote/application-note-archives

apparently this changed between version 2.0 and 4.5.0

would it be possible to use the ZipFile API, since it apparently doesn't suffer from the bug? i'm not familiar with it.

sebkur:
It would be possible to use ZipFile, yes, we would need to copy the file to a temporary location though instead of just an in-memory byte[], because the ZipFile API only works with File arguments. Could be OK I guess

sebkur:
Just filed a bug on the commons-compress JIRA, let's see whether somebody reacts in the near future: https://issues.apache.org/jira/projects/COMPRESS/issues/COMPRESS-549?filter=allopenissues

sebkur:
See this for an example of how to copy the input stream to a temporary location and opening the file using ZipFile:
https://github.com/sebkur/test-zip-impls/blob/master/src/test/java/de/topobyte/zip/tests/TestZipFile.java

sebkur:
although... hmm... OdfPackage has a constructor with File anyways, and apparently it used to have a separate method for initializing the ZipHelper class using the File argument which then relies on ZipFile internally. https://github.com/tdf/odftoolkit/blob/cb285ca81654e75620664084a3f5048d5b569c82/odfdom/src/main/java/org/odftoolkit/odfdom/pkg/OdfPackage.java#L522 it's just commented. In case of OdfPackage being constructed from file, we would not need to copy to a temporary file, in case of OdfPackage being constructed from InputStream, we would need to copy to a temporary location

mistmist:
oh so it can already use ZipFile (almost), nice find...

sebkur:
OK, turns out I was wrong concerning Commons Compress, which can read the ODT file just fine.
I misunderstood the values for compressed size and uncompressed size of -1 which actually is a code for "unknown size". The data can be read without any problems from the InputStream however. This totally makes sense as the "data descriptor" flag in PKZip mandates that the CRC and size values of the file header must be set to zero, hence the library reading the ZIP archive can't possibly know the size of the upcoming file. It has to start decoding until the stream hits a natural end with respect to the compression method. I changed my testing project (https://github.com/sebkur/test-zip-impls/commit/7e257f00b238c38a5e19740e4f787f10b56c42d5) and the test for Commons Compress passes now.

Under this light, I think we can re-evaluate whether it makes sense to use this library!

sebkur:
I'll continue work on this in a new draft pull request #58

PR #57, opened 2020-07-15
Remove @deprecated annotations from original API - fixes #56

### LINKED ISSUE #56: Undeprecate original API
As I understand it, the deprecation warnings in 0.9.0-RC1 for the original API should be ignored since the Simple API is going away. 

Would it be possible to remove the deprecation tags so that developers aren't misled into doing a pointless migration to an API that is already dead?

tfmorris:
Since the recommendation to migrate to the (to be removed) Simple API is misleading, remove the deprecation annotations from the original API.

Fixes #56

### DIFF
--- a/odfdom/src/main/java/org/odftoolkit/odfdom/doc/OdfChartDocument.java
+++ b/odfdom/src/main/java/org/odftoolkit/odfdom/doc/OdfChartDocument.java
@@ -32,8 +32,6 @@ import org.xml.sax.SAXException;
 /**
  * This class represents an empty ODF document , which will be in general embedded
  * in an existing ODF (Spreadsheet) document.
- *
- * @deprecated As of release 0.8.8, replaced by {@link org.odftoolkit.simple.ChartDocument} in Simple API.
  */
 public class OdfChartDocument extends OdfDocument {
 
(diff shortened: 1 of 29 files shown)


mistmist:
the content of the commit is fine - i'm just scratching my head which branch it should be merged into... probably not 0.0-ODF1.3 since it has nothing to do with ODF 1.3, but instead the 0.9 branch, which confusingly enough is called "master" - we really should clean that up some time...

i can't change the target branch in the web ui, are you able to change it tfmorris?

tfmorris:
> the 0.9 branch, which confusingly enough is called "master" 

Yes, **super** confusing since the `master` normally represents the very latest development, so in the odftoolkit case, if I understand the release plan, it would be 2.x.

> i can't change the target branch in the web ui, are you able to change it tfmorris?

It's more than just changing the target branch, since the contents of the two branches are different and it doesn't look like the `master` branch actually has the `@deprecated` tags.

Since I clearly have no idea how things are structured, I'm just going to close this and let you folks figure it out.

svanteschubert:
Yes, sorry for the confusion. 
The master will become 0.9.0 and the 1.0.0_SNAPSHOT will be the 1.0.0 where future development will be placed.
All related to some forks of companies in the past. I explained the history recently: 
https://listarchives.odftoolkit.org/dev/msg00023.html
I will be in Denmark on vacation to the end of the month, releases of 0.9.0 and 1.0.0 will happen afterwards.
The new second ODF 1.3 committee specification release and the new GitHub with tooling took its time: 
https://github.com/oasis-tcs/odf-tc

Still thanks for the effort, Tom!
Svante

mistmist:
okay i was so annoyed by this now that i deleted the master branch and replaced it with "0.9" branch.

PR #58, opened 2020-07-25
Commons compress

sebkur:
This draft exists to continue the work proposed and discussed in #53


sebkur:
On the old PR, the Travis pipeline failed because in https://github.com/tdf/odftoolkit/pull/53/commits/380d99c5c9c3e0c3956a6aba99f06df2d76a7936 I deleted the duplicate class OdfXMLHelper from src/main instead of src/test. On this branch I deleted the right file from the testing sources.

sebkur:
To be clear the actual migration to commons-compress is not yet in here, this is just a draft

sebkur:
Okay, nice everything works again after switching to commons compress in c1f18a50c8c12fafa81df525803089e18e71974d

sebkur:
Time to add the additional test case with new file `gpg13.odt` from #48

sebkur:
Hmm, okay, that is strange. Travis passes all checks however, if I run `InvalidPackageTest` from the `odfvalidator` module from within Eclipse, 4 out of 8 tests actually fail: validateDocumentSignature, validateEncryptedODT, validatePackage3 and validateEncryptedODF13WithInvalidOpenPGP. I can also reproduce it on the command line using

    mvn -Dtest=InvalidPackageTest -DfailIfNoTests=false test

on a Linux Mint 19.1 machine, which is based on Ubuntu Bionic, using OpenJDK 11.0.7

sebkur:
Could it be that Travis is not running all the tests?

mistmist:
> Could it be that Travis is not running all the tests?

indeed there's no output indicating the validator tests are actually run;

[INFO] --- maven-compiler-plugin:3.8.1:testCompile (default-testCompile) @ odfvalidator ---
[INFO] Nothing to compile - all classes are up to date
[INFO] 
[INFO] --- maven-surefire-plugin:2.22.2:test (default-test) @ odfvalidator ---
[INFO] ------------------------------------------------------------------------

while lots of test output following these:

[INFO] --- maven-surefire-plugin:2.19.1:test (default-test) @ schema2template ---

[INFO] --- maven-surefire-plugin:2.22.2:test (default-test) @ odfdom-java ---

sebkur:
Just rebased this on [sebkur:improve-travis](https://github.com/sebkur/odftoolkit/tree/improve-travis) to see what happens with the updated Travis config

sebkur:
Great now Travis is executing the tests, which are failing, which is good, because we can now identify and fix the actual problems with the migration to commons compress :)

sebkur:
Now that the tests get executed by Travis, going back to the state before adding the new test cases. Before we try new stuff using commons compress, let's get the old stuff working.

sebkur:
Now the tests really pass. Time to add changes from #48

sebkur:
Okay, finally, this branch has arrived at the point where we can actually work on the new test case from #48. Currently the build fails with InvalidPackageTest (see https://travis-ci.com/github/tdf/odftoolkit/builds/177314760#L7963) @mistmist can you take a look at this? Thanks.

sebkur:
There is an error parsing the manifest.xml

    gpg13.odt/META-INF/manifest.xml[2,194]:  Error: unexpected attribute "manifest:version"
    s:experimental:office:xmlns:loext:1.0"
                                     ----^

mistmist:
> There is an error parsing the manifest.xml
> 
> ```
> gpg13.odt/META-INF/manifest.xml[2,194]:  Error: unexpected attribute "manifest:version"
> s:experimental:office:xmlns:loext:1.0"

the problem is:

internal:/schema/odf1.1/OpenDocument-manifest-schema-v1.1.rng:  Info: parsed.

it should be using 1.3 schemas, not 1.1 schemas.

there's some version autodetect code somewhere but iirc it doesn't work for encrypted documents... anyway i wrote the getVersionFromManifest function for this purpose, it's possible that you need to extract it from my commit.

sebkur:
hmm, so #48 really does two things?
1. work around the buggy java.util.zip package
2. implement some magic manifest handling

why do we need 2.? Is LibreOffice creating strange manifest.xml files for encrypted documents?

mistmist:
the reason is that ODFDOM was unable to even load an encrypted ODF package due to the zip bug, so we had to find a way to get at the manifest.xml and autodetect the ODF version from it without ODFDOM's help.

odfdom/src/main/java/org/odftoolkit/odfdom/pkg/ZipHelper.java
```diff
@@ -95,7 +95,7 @@ String entriesToMap(Map<String, ZipArchiveEntry> zipEntries) throws IOException,
             }
             // JDK 6 -- the try/catch is workaround for a specific JDK 5 only problem
             if (!e.getMessage().contains("missing entry name")
-                && !System.getProperty("Java.version").equals("1.5.0")) {
+                && !"1.5.0".equals(System.getProperty("Java.version"))) {
```
mistmist, line 103: this is dead code anyway, we require, i forgot, certainly at least java 8 - easiest to just delete the whole thing

sebkur:
yes, okay, but assuming that we are now able to load the package because we are not suffering from the zip bug anymore, where can we hook into the existing code to add that additional validation?

odfdom/src/test/java/org/odftoolkit/odfdom/pkg/PackageTest.java
```diff
@@ -343,6 +343,7 @@ public void validationTest() {
     Map expectedErrors3 = new HashMap();
     expectedErrors3.put(OdfPackageConstraint.MANIFEST_NOT_IN_PACKAGE, 1);
     expectedErrors3.put(OdfPackageConstraint.MIMETYPE_WITHOUT_MANIFEST_MEDIATYPE, 1);
+    expectedErrors3.put(OdfPackageConstraint.MIMETYPE_HAS_EXTRA_FIELD, 1);
```
mistmist: nice!

sebkur:
Actually I think for a clear separation of concerns, we should split the two issues. One issue is working around the limitations of java.util.zip, the other is implementing handling of encrypted files and their validation. The first seems to be solved after commit f40c756 where all tests pass, i.e. commons compress handles everything implemented hitherto fine, all checks pass. If we agree that is makes sense to work around the limitations of java.util.zip by building on a solid foundation (commons compress) we can merge this and start working on the second issue, namely extending the validation mechanism so that they also work with encrypted ODF packages.

validator/src/test/java/org/odftoolkit/odfvalidator/InvalidPackageTest.java
```diff
@@ -157,10 +157,6 @@ public void validateEncryptedODT() {
       t.printStackTrace(new PrintWriter(errors));
       Assert.fail(t.toString() + "\n" + errors.toString());
     }
-    Assert.assertTrue(output.contains("Fatal: only DEFLATED entries can have EXT descriptor"));
-    Assert.assertTrue(
-        output.contains(
-            "Fatal: The document is encrypted. Validation of encrypted documents is not supported."));
```
mistmist: we can only validate the manifest.xml in encrypted files; i think this should be a warning, maybe a fatal error is a bit too much.

sebkur:
Concerning the second issue, from what I get from the codebase, I believe probably the right position to start fixing the validation of the encrypted package's manifest.xml file is in ODFRootPackageValidator's validateManifest() method, where the manifest gets validated. I that validation fails, we can have that fallback validation getting called.

odfdom/src/main/java/org/odftoolkit/odfdom/pkg/ZipHelper.java
sebkur: It's not dead, fixed it because I ran into the NPE

validator/src/test/java/org/odftoolkit/odfvalidator/InvalidPackageTest.java
sebkur: The got removed in 60993ebd75e5cf54e643e097d because this does not happen anymore

sebkur:
@mistmist you're looking at outdated code because some of the commits I made I reverted in later commits, maybe look at the summarized changes here: https://github.com/tdf/odftoolkit/compare/master...sebkur:commons-compress

odfdom/src/main/java/org/odftoolkit/odfdom/pkg/ZipHelper.java
mistmist: well if you want to get pedantic about it, there is no dead code but an always-false condition, and i'm arguing that going from {false, ⊥} to {false} is not as much an improvement as just removing the condition.

sebkur: I'm not being pedantic. it's not always false. on my setup System.getProperty("Java.version") returns null raising an NPE

sebkur: The code is not needed anymore, that I can agree on

sebkur: Seems to be an invalid assumption to rely on the system property "Java.version" being set

sebkur: hmm, turns out system properties are not case-insensitive. "java.version" would work, "Java.version" doesn't

mistmist:
so the version problem is that ODFPackageValidator::getVersion only looks at root document files but not manifest.xml; in encrypted documents the root document files are encrypted so that doesn't work.

presumably ODFPackageValidator::_validate is what validates the document streams - probably you'd get some sort of exception for an encrypted document because XML parser won't be happy to see it, that would need to be handled somehow, perhaps result in a warning.

perhaps it would be best to skip encrypted entries outright, the manifest.xml contains the info which entries are encrypted and i hope ODFDOM makes it available.

validator/src/test/java/org/odftoolkit/odfvalidator/InvalidPackageTest.java
mistmist: hmm i think there should be some output that the document streams weren't actually validated though.

sebkur:
I think it makes sense to first determine if the file is encrypted or not and then call different validation mechanism for both cases

sebkur:
Let's continue the discussion in #61

sebkur:
and for the migration to commons compress please refer to #60

PR #59, opened 2020-07-27
Let Travis run the tests

sebkur:
Let's see if this helps with #58

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,3 +11,5 @@ jdk:
   #- oraclejdk11
   #- openjdk8
   - openjdk11
+
+script: mvn clean install -Ppedantic


sebkur:
Well, that looks good to me. Build time increased from 6 Minutes to 10, which makes sense. Also the output is more verbose now.

sebkur:
I have read this: https://docs.travis-ci.com/user/languages/java/#projects-using-maven
which says that without further configuration Travis does the following.

Install dependencies:

    mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V

Run the default script command:

    mvn test -B

In order to change that behavior, we can change the script command, which I did. No idea why `mvn test -B` is not enough, but Svante mentioned that he usually runs `mvn clean install -Ppedantic` for testing, so this is what I put there as a script command.

mistmist:
looks much easier than trying to understand what maven does.
all i could find out is that if i remove this

                <configuration>
                    <excludes>
                        <exclude>**/*.java</exclude>
                    </excludes>
                </configuration>

below 
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-surefire-plugin</artifactId>
                <version>2.22.2</version>

then "mvn test" will run one test validator/src/test/java/org/odftoolkit/odfvalidator/jar/ITJarTest.java - which fails - but won't run any of the test in validator/src/test/java/org/odftoolkit/odfvalidator/*.java

PR #60, opened 2020-07-28
Commons compress only

sebkur:
Extracted from #58 
This PR introduces a migration from java.util.zip.* to commons compress


mistmist:
okay no feedback from Svante so lets just merge this

PR #66, opened 2020-09-11
validator: work around mysterious behavior of maven-assembly-plugin

mistmist:
The problem is that somehow directories in dependency jars become empty
files in the -jar-with-dependencies jars.  Not an issue when invoked via
java -jar, but Guilhem reports that this prevents the web app from
starting.

Good:

       0  Stored        0   0% 05-26-2014 21:18 00000000  META-INF/maven/commons-beanutils/

Bad:

       0  Defl:N        2   0% 05-26-2014 21:18 00000000  META-INF/maven/commons-beanutils

So the affected directories were found with
> unzip -v validator/target/odfvalidator-1.0.0-BETA2-SNAPSHOT-jar-with-dependencies.jar | grep "Defl:N *2 "

The jars in odfdom, xslt-runner are also affected.

Bisecting shows this started with:

commit 2a6d4fc2b45eed57f83604c5de92941629508ece

    Merging the ODF collaboration work from https://github.com/svanteschubert/odftoolkit/tree/odf-changes

... and bisecting in that git repo gives:

commit b04c5828ba32987adde85396d3ec82374c255124

    Merge of collaboration feature branch with main trunc. Becoming 1.0.0-SNAPSHOT of the ODF Toolkit at The Document Foundation (TDF)

... but it's not obvious what change caused this.

The only obvious work-around is not actually obvious but at least it's
quite undeniably ridiculous, and i disclaim any insight of why it
appears to give the correct result.

Note: i hope somebody has a better idea than this nonsense? Maybe @sebkur :)

### DIFF
--- a/validator/src/main/assembly/src.xml
+++ b/validator/src/main/assembly/src.xml
@@ -28,6 +28,37 @@
             <outputDirectory></outputDirectory>
             <useProjectArtifact>true</useProjectArtifact>
             <unpack>true</unpack>
+            <!-- HACK: this suppresses empty files of that name, will get the directories that we want instead
+              FIXME: WHY DOES THIS WORK???
+            -->
+            <unpackOptions>
+                <excludes>
+                    <exclude>META-INF/maven/commons-beanutils</exclude>
+                    <exclude>META-INF/maven/commons-beanutils/commons-beanutils</exclude>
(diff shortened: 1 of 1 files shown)


svanteschubert:
Fixed this with Michael in Mathml01 PR

PR #68, opened 2020-09-17
validator: also resurrect MathML 1.01 DTD

mistmist:
This is referenced from ODF 1.0/1.1 configurations.


svanteschubert:
First I have added both MathML schema (MathML1.1 DTD and MathML2 W3C schema) to the paths of the properties files.
But still, get these errors.. strange...

I have tested the ODF 1.3 part 4 document from our ODF-TC GitHub
https://github.com/oasis-tcs/odf-tc/blob/master/src/main/resources/odf1.3/OpenDocument-v1.3-part4-formula.odt

by using the validator WAR created from this branch in Tomcat (9.0.17) and still get this error:
OpenDocument-v1.3-part4-formula.odt/Object 337/content.xml[2,6]: Error: Document is invalid: no grammar found.
OpenDocument-v1.3-part4-formula.odt/Object 337/content.xml[2,6]: Error: Document root element "math", must match DOCTYPE root "null".

svanteschubert (approved):
Thanks, Michael. I have reviewed and added three fixes of related issues along on the road...

validator/src/main/java/org/odftoolkit/odfvalidator/Main.java
```diff
@@ -213,7 +214,7 @@ public static void main(String[] aArgs) {
         aConfig.setProperty(Configuration.DSIG_SCHEMA, aDSigSchemaFileName);
       }
       if (aMathMLSchemaFileName != null) {
-        aConfig.setProperty(Configuration.MATHML2_SCHEMA, aMathMLSchemaFileName);
+        aConfig.setProperty(Configuration.MATHML3_SCHEMA, aMathMLSchemaFileName);
```
mistmist, line 217: this breaks the build, but it's fixed in the next commit

pom.xml
```diff
@@ -233,7 +233,11 @@
             <plugin>
                 <groupId>com.cosium.code</groupId>
                 <artifactId>git-code-format-maven-plugin</artifactId>
-                <version>2.5</version>
+                <version>2.6</version>
```
mistmist, line 236: this change isn't related to what's in the commit message

validator/pom.xml
```diff
@@ -94,23 +94,30 @@
             </plugin>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-jar-plugin</artifactId>
-                <version>3.1.1</version>
+                <artifactId>maven-war-plugin</artifactId>
```
mistmist, line 97: this change isn't related to what's in the commit message

validator/pom.xml
```diff
@@ -175,18 +182,33 @@
                 <artifactId>maven-assembly-plugin</artifactId>
                 <configuration>
                     <archive>
+                        <index>true</index>
+                        <manifest>
+                            <Main-Class>org.odftoolkit.odfvalidator.Main</Main-Class>
```
mistmist, line 187: this change isn't related to what's in the commit message

validator/src/main/resources/schema/mathml3/mathml3-common.rng
```diff
@@ -0,0 +1,257 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+      This is the Mathematical Markup Language (MathML) 3.0, an XML
```
mistmist, line 3: probably we need to add some copyright blurb to the NOTICE file, it already has something for MathML 1 and 2

validator/src/main/resources/schema/odf1_3.properties
```diff
@@ -25,10 +25,8 @@
...
 <entry key="strict-schema"></entry>
 <entry key="schema">/schema/odf1.3/OpenDocument-schema-v1.3.rng</entry>
-<entry key="mathml-schema">/schema/mathml1.01/math.dtd</entry>
-<entry key="mathml2-schema">/schema/mathml2/mathml2.xsd</entry>
-<entry key="dsig-schema">/schema/odf1.3/OpenDocument-dsig-schema-v1.3.rng</entry>
-<entry key="path1"/>
-<entry key="filter"/>
+<entry key="mathml1.01-schema">/schema/mathml1.01/math.dtd</entry>
+<entry key="mathml3-schema">/schema/mathml3/mathml3.rng</entry>
+<entry key="dsig-schema">/schema/odf1.3/OpenDocument-dsig-schema-v1.3.rng</entry>
```
mistmist, line 30: trailing whitespace was added to this line

validator/src/main/java/org/odftoolkit/odfvalidator/Main.java
svanteschubert: as I said, I split the 3 commits after fixing alls and very roughly.. ;-)

pom.xml
svanteschubert: Yes, as soon GitHub comes with the tooling to split commits, I am on it doing this little things... (or you pay me by hour)  ;-)

validator/pom.xml
svanteschubert: If this should be a question of yours, the manifest of the WAR file can only be adjusted by the renamed plugin.
I desired to trigger the WAR by command line similar to the JAR.

validator/pom.xml
svanteschubert: Do not be so picky, it is the same as the previous, starting the WAR by JAR command line parameter

validator/src/main/resources/schema/odf1_3.properties
svanteschubert: We should fix this in the root pom.xml where whitespacehandling is handled.
I fix this...

svanteschubert (approved):
Thanks for the feedback and fixes of notice & whitespace

PR #79, opened 2020-11-30
Odfdom osgi bundle

### LINKED ISSUE #78: Make odfdom-java an OSGi bundle
We have prepared an internal OSGi bundle for odfdom-0.8.10 and we would like to use odf-dom-1.0.0 as an OSGi bundle.
We already have a prototype running and encountered, that the crucial point is, that the java-rdfa library last released in 2012 is dependent on jena-2, which incurs a lot of historic dependencies.
I have finished the modernization of the library in the repository https://github.com/iteggmbh/java-rdfa and I'm currently discussing with Sonatype on how to gain control over the deployment to the net.rootdev groupId.
If the library needs a change in the groupId, it might be a feasible solution to change to groupId to org.odftoolkit, so the odftoolkit team might deloy a cleaned up version of java-rdfa.

wglas85:
This branch builds against the new java-rdfa-1.0.0-BETA1 release prepared by us.
Thus, it cleans up many dependency problems.
Additionally, it add OSGi bundle headers to the oodfdom jar file, which fixes #78

### DIFF
--- a/odfdom/pom.xml
+++ b/odfdom/pom.xml
@@ -165,6 +165,7 @@
                 <configuration>
                     <archive>
                         <index>true</index>
+                        <manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
                         <manifest>
                             <mainClass>org.odftoolkit.odfdom.JarManifest</mainClass>
                         </manifest>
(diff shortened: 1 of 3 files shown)


svanteschubert (approved):
Thank you for the contribution, I reviewed your changes and was able to build your branch with "mvn install -Ppedantic"

PR #80, opened 2021-01-18
Adds a Pattern constructor to TextNavigation

gtache:
The new constructor is necessary if we want to use Pattern flags (like unicode support).     
Compiling the pattern only once should also improve performance a bit.     
(Also cleans up the file a bit)

By the way, the .editorConfig file specifies 4 spaces for indentation, but the file is indented with 2 spaces. Should it be reformatted?

Thanks for the toolkit by the way, I've used it quite a bit already for manipulating ods and odt, and it works (most of the time) like a charm!

### DIFF
--- a/odfdom/src/main/java/org/odftoolkit/odfdom/incubator/search/TextNavigation.java
+++ b/odfdom/src/main/java/org/odftoolkit/odfdom/incubator/search/TextNavigation.java
@@ -37,7 +37,7 @@ import org.w3c.dom.NodeList;
 public class TextNavigation extends Navigation {
 
   private String mMatchedElementName = "text:p,text:h";
-  private String mPatternText;
+  private Pattern mPattern;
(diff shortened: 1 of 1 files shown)


svanteschubert:
Thanks for the patch and sorry for the delay in answering.

Indeed the format plugin is not working perfect, seems I need to add/commit changes before they are being considered to be formatted automatically  (missing feature/bug).
To format manually you migth run by commandline:
mvn git-code-format:format-code -D"gcf.globPattern=**/*"
or 
mvn git-code-format:format-code -Dgcf.globPattern=**/*

After this manual format, your patch builds fine. 
I will merge and fix it right away to save us some time!

Thanks again!
Svante

svanteschubert (approved):

gtache:
No problem, thanks for the merge and for the command to format!

PR #89, opened 2021-05-19
validator: put git revision on main page of webapp

mistmist:
So that it's obvious to see what is deployed on odfvalidator.org.

Also print it in jar file when invoked with -V, because why not.

Use git-commit-id-plugin maven plugin for this.


svanteschubert (approved):
Thanks for the patch, Michael!

PR #94, opened 2021-06-09
Odftoolkit 187 1.0 odf1.3

mistmist:


mistmist:
the 2nd commit is mostly for testing, not sure if we want to merge this - ideally those assertions should be in tests?

svanteschubert (approved):
Great work, Michael! You fixed a more than 10 years old issue written 3 months before the downfall of StarOffice! 

Thousand thanks! 
Svante

Issue #54, opened 2020-06-30
Enabling Java 9 modularity


svanteschubert:
Someone who is already using Java 9 modularity might assist us in adding this concept to the ODF Toolkit JARs.

See also:
https://www.oracle.com/corporate/features/understanding-java-9-modules.html
https://www.baeldung.com/java-9-modularity

miasma:
At least for my simple use case, adding the following `src/main/java/module-info.java` seemed to work:
```java
module org.odftoolkit.odfdom {
  requires java.desktop;
  requires java.logging;
  requires java.xml;
  requires java.rdfa;
  requires xercesImpl;
  requires org.json;
  requires org.slf4j;
  requires org.apache.jena.core;
  requires org.apache.commons.compress;
  requires org.apache.commons.lang3;
  requires commons.validator;
  requires serializer;

  exports org.odftoolkit.odfdom;
// and other exported packages
}
```

I had to disable the tests `java/org/odftoolkit/odfdom/pkg/GRDDLTest.java` and `java/org/odftoolkit/odfdom/pkg/RDFMetadataTest.java` since I couldn't figure out how to make the jena-core:tests visible. While this is nice to have, it still won't work properly until [xercesImpl-2.12.0.jar, serializer-2.7.2.jar, java-rdfa-1.0.0-BETA1.jar, commons-validator-1.6.jar] have also migrated.

svanteschubert:
Hello Miasma,

Thank you for your contribution and insights. 
So all our dependent project JARS you mentioned need to be transitioned to JDK 9 prior as well.

1. xercesImpl-2.12.0.jar &  serializer-2.7.2.jar [mentioned to be working](http://apache-xml-project.6118.n7.nabble.com/Xerces-J-2-On-JDK-9-td44032.html)
2. commons-validator-1.6.jar [release notes for 1.7 mention a breaking change due to JDK 9 local usage - indicated they have JDK9 on the radar - I have not tested, but optimistic](https://commons.apache.org/proper/commons-validator/changes-report.html)
3. No insight on Jena, likely we have to test the latest version and/or build locally with JDK 9 enabled and contribute it back
4. java-rdfa-1.0.0-BETA1.jar is from Damian Steer, who I once met. I just have written him about JDK 9 modules and the repo of
https://repo1.maven.org/maven2/net/rootdev/java-rdfa/1.0.0-BETA1/
as the former repo seems still very old:
https://github.com/shellac/java-rdfa

[...]

PR #132, opened 2021-11-13
Remove dependency to xml-apis:xml-apis.

### LINKED ISSUE #131: Remove dependency to xml-apis
Now that this project requires Java >= 11, we should consider removing the dependency to `xml-apis:xml-apis` whose contents are already provided by the JRE for those versions of Java (I believe). It would help solving issues related to Java 9 modularity (#54) since those classes should not be exposed by two different modules.

wetneb:
No longer needed for Java 11 and above.
Also solves compilation problems for users who enforce
Java 9 modular compilation (https://github.com/tdf/odftoolkit/issues/54#issuecomment-968061438)

Closes #131.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -60,13 +60,6 @@
     <dependencyManagement>
         <dependencies>
             <dependency>
-                <groupId>xml-apis</groupId>
-                <artifactId>xml-apis</artifactId>
-                <!-- never use version 2.0.2 as it is prior version 1.4.01 -->
-                <!-- see https://mvnrepository.com/artifact/xml-apis/xml-apis -->
-                <version>1.4.01</version>
-            </dependency>
-            <dependency>
                 <groupId>xerces</groupId>
                 <artifactId>xercesImpl</artifactId>
                 <version>2.12.1</version>


wetneb:
@svanteschubert any chances this could make it into the next release? I think it would save users some trouble :)

svanteschubert (approved):
Good Catch! 
I have not read nor heard anything about that problem and its easy solution! Well done!

You only removed the xml-apis and it builds fine under Linux:

        git fetch origin pull/132/head:131-xml-apis
        git checkout 131-xml-apis
        git clean install

see https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/checking-out-pull-requests-locally

mistmist:
README.md and docs/ReleaseNotes.html say JDK 9 is required; does this mean that JDK11 is required now?

this would be fine with me, hopefully not many still use 9 or 10.

svanteschubert:
Yes, it seems fine to me that we test and implement against the major used LTS version (JDK 11 for now) ->
https://www.oracle.com/java/technologies/java-se-support-roadmap.html
(Unless we are very eager to use JDK 17 feature at some time in the future) ;-)

In addition, I quickly installed JDK 9 and tested it, we are already setting JDK Version to 11:
https://github.com/tdf/odftoolkit/blob/master/pom.xml#L37

Good reminder to update our documentation that we tested against JDK 11 to our next release! 👍 
(next LTS JDK 17 should build & work nevertheless)...

PR #134, opened 2021-11-13
Add Continuous Integration with GitHub Actions.

### LINKED ISSUE #133: Continuous integration to run tests on pull requests
This repository does not seem to have Continuous Integration set up.
As a contributor, when I open a pull request to a Java library, I expect that tests will be run automatically on it and that the maintainer will be able to merge this pull request based on those results and their review.

It would be easy to add this using GitHub Actions.

wetneb:
Closes #133.

This is a stock Maven CI configuration suggested by GitHub, we can improve it later on (such as running tests on multiple versions of Java, for instance 11 and 17).


wetneb:
Note that we cannot really see the result of this PR before it is merged, sadly. I would encourage to try it out and revert if you are not happy @svanteschubert  :)

mistmist:
so i guess since Travis CI went MIA, we can try this instead :)

we would also need it to work on PRs for the 0.9 branch, does it read the file always from the master branch or does it check every branch for the file?

wetneb:
I think it only reads the configuration files from the default branch of the repository (but it might not be true for all action types). Adding the `0.9` branch to the config, then.

svanteschubert:
That is pretty cool!

But I must admit, I am not really sure how to test this, so I just read the lines 

        mvn -B package --file pom.xml

and called 'mvn -h' on the command line to check for the parameters 

        -B,--batch-mode     Run in non-interactive (batch)
                            mode (disables output color)

        -f,--file           Force the use of an alternate POM
                            file (or directory with pom.xml)

I assume this POM file is our project pom file and alternate to the internal GitHub builds.
As I am a newbie on GitHub actions, I can not tell if the path is correct. I assume pom.xml is expected at our project root, yes?

One last thing, we might want to add the parameter **-Ppedantic** to run the Apache Rat plugin checking for missing copyright header, etc.
Also, wonder if we should not add "clean" and "install" in addition like we did in Travis.

Just awaiting your reply (and/or update) and then I will wave it through and see how it works ;-)

wetneb:
Yeah I am not sure why `--file` is specified in this default configuration file, since by default Maven already looks at the `pom.xml` file in the current directory anyway. It would also work without I assume.

I added the `-Ppedantic` flag.

svanteschubert (approved):
Now let's give it a try! :-)

and thank you very much for the patch, Antonin!👍

wetneb:
Great! It seems to have worked: https://github.com/tdf/odftoolkit/runs/4216261903?check_suite_focus=true

PR #136, opened 2021-11-18
Fixes TextSelection delete in case of text:s

gtache:
Consider the following (simplified) example:

`My name is Henri <text:s></text:s>Smith and this is Mary Poppins`

Currently, if we have a selection on `Henri Smith` and we replace it with `A`, we'll get `My name is A<text:s c='1'></text:s> and this is Mary Poppins`, so we have an unwanted double space between `A` and `and`.
Worse, if we have a selection on e.g. `Mary Poppins` at the time and we want to replace it after replacing `Henri Smith`, we get an exception because the expected index is off by one.      
The solution seems to be to simply delete the `<text:s>` node instead of keeping it.

Let me know if there's something I've overlooked, in case the solution is not as trivial as what I thought.


svanteschubert:
@gtache 
Hej Guillaume, many thanks for your patch! I must admit I have overseen it somehow before the release so it did not make it, but I need to finish the source generator soon for ODF 1.2, so there will be a follow-up release to integrate it!

Can you do me a favor and provide a test case with the fix in the existing test class:
https://github.com/tdf/odftoolkit/blob/master/odfdom/src/test/java/org/odftoolkit/odfdom/incubator/search/TextSelectionTest.java
Which would fail prior to your patch and succeed afterward?
This would guarantee that this feature will not break away again! :-)

Thanks again and have a great weekend!
Svante

gtache:
Hello,     
No problem, I added the test as requested and confirmed it wasn't working on the current master (an InvalidNavigationException is thrown).


Have a nice day,        
Guillaume

svanteschubert:
## How To checkout PR Branches locally

I realized (refound) I could test a Pull Request by locally update and checkout via:
 git fetch origin pull/**ID**/head:**BRANCHNAME**
see https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/checking-out-pull-requests-locally

I believe the branch name is arbitrary but to be consistent I have reused yours:
 git fetch origin pull/136/head:deleteS


## State of Review and Suggestions

Your branch builds fine with 'mvn clean install -pPedantic' and your test document and test look very professional (lambda not yet in my blood, have to do the homework).. ;-)
And even better, if I undo your fixes manually, your new test fails! 👍 

You have added to the existing document the following, which was a good idea:
<text:p text:style-name="P6">This is a paragraph containing multiple <text:s text:c="2"/>spaces between some <text:s/>words, <text:span text:style-name="T23">and not between some others</text:span>.</text:p>

If you take open the document in LibreOffice and/or look to the updated regression test of JSON changes you see both, that there are three spaces between 'multiple' and 'spaces': in XML 'multiple <text:s text:c="2"/>spaces'.

I would like to ask you two things to think about:

1. Imagine the test ODT would include "multiple <text:s text:c="5"/>spaces" and now we delete 'ple   ' => in other words the prefix 'ple' plus three spaces after 'ple'! In this case it might become "multi<text:s text:c="3"/>spaces", right?
For <text:s> being 'space' element with a counter c, see https://docs.oasis-open.org/office/OpenDocument/v1.3/os/part3-schema/OpenDocument-v1.3-os-part3-schema.html#element-text_s
Guess we need the prior and your part depending on if the deleted string ends within a long space block.
2. Search over the IBM donated code (yeah, but blame me that I have not realized it earlier) for comment "if text:s?????????". The part you fixed does exist as duplicated code three times - likely all parts need the same fix, right? (you might remove the comment afterwards) ;-) 

I am (again) under time pressure, otherwise, I would have assisted you more with the fix. At least I try to document it now for everyone! Software is an iterative process! Thanks for your help so far and keep up the good work, Guillaume!  :-)

mistmist:
i hope i've fixed the outstanding problem here...
unfortunately github doesn't allow to push to a pull-request ref, so i had to create a new PR at
https://github.com/tdf/odftoolkit/pull/154
=> closing this one

PR #140, opened 2021-11-27
FIXING: Logging error message, when executing ODFDOM JAR by command line

svanteschubert:
You some test ODT, for instance, https://tdf.github.io/odftoolkit/docs/presentations/character-styles.odt
And use the latest release JAR (with all dependencies), you may find here:
https://repo1.maven.org/maven2/org/odftoolkit/odfdom-java/0.10.0/odfdom-java-0.10.0-jar-with-dependencies.jar

java -jar .\odfdom-java-0.10.0-jar-with-dependencies.jar character-styles.odt

The following error message will be displayed ahead of the ODT expressed as change operations:
_SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details._

Applying the fix described on [StackOverflow](https://stackoverflow.com/questions/7421612/slf4j-failed-to-load-class-org-slf4j-impl-staticloggerbinder) worked.

Any comments or improvements?

### DIFF
--- a/docs/website-development.html
+++ b/docs/website-development.html
@@ -96,13 +96,12 @@
       <h2 id="building-javadoc-and-publishing-the-javadoc">Building JavaDoc and publishing the JavaDoc</h2> 
       <p>JavaDoc is generated as part of by calling from the root of the project the command:</p> 
       <div class="codehilite">
-        <pre><code class="language-shell">mvn clean install
+        <pre><code class="language-shell">(mvn install &amp;&amp; cd odfdom &amp;&amp; mvn javadoc:javadoc)
 </code></pre>
       </div> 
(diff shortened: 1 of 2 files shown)


mistmist (changes requested):
this is wrong and i've just fixed the source of it with commit 51f415b28f476b807e526f841e53c2efcd68ae67

mistmist:
this was pushed manually as commit beb5a56c404f2f62e4454c8f083a76b7767778c6

svanteschubert:
@mistmist 
Thanks, Michael! No idea how the other file commit went into this PR - it was a very very slow push and just happened before the 2hour GitHub outage... might have something to do with that.. anyway happy it's in now! :-) 👍

PR #147, opened 2022-01-03
Automated GitHub Deployment by GitHub Action

svanteschubert:
I have started looking into GitHub actions over the holidays and created GitHub action that makes an automatic GitHub deployment whenever a Git tag is being added. 

For example:
- git tag -sm 0.11.0-SNAPSHOT 0.11.0-SNAPSHOT
- git push --follow-tags

results to https://github.com/svanteschubert/odftoolkit/releases/tag/odftoolkit-0.11.0-SNAPSHOT
(Note: I used a test tag 0.10.11 instead of 0.11.0-SNAPSHOT, but the version is being extracted in the action from the pom)

If someone does not like the commit message as part of the title (like me), it still seems an easy manual tweak - as I have not found out how this is handled by the plugin https://github.com/svenstaro/upload-release-action I am using. ;-)
There was once an official GitHub plugin, but this is for some unknown reason no longer supported: https://github.com/actions/create-release

I have not completely wrapped my head around their actions concepts, for instance, I am not able to limit the events, currently, it is building the toolkit sometimes twice, once for the push event and once for the tag event, but only skimmed their documentation: https://docs.github.com/en/actions/quickstart

There is also a way to do an automated maven release see https://blog.frankel.ch/github-actions-maven-releases/
Might try such action for the next release for some special tag :-)

### DIFF
--- a/.github/workflows/deployment.yml
+++ b/.github/workflows/deployment.yml
@@ -1,0 +1,39 @@
+name: Publish JAR on tagging
+
+on:
+  push:
+    tags:
+      - '*'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up JDK 11
+      uses: actions/setup-java@v2
+      with:
+        java-version: '11'
+        distribution: 'adopt'
+        cache: maven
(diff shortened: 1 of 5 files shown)


mistmist (approved):
lgtm

PR #169, opened 2022-04-29
[Snyk] Upgrade org.apache.tinkerpop:tinkergraph-gremlin from 3.5.2 to 3.6.0

svanteschubert:
<p>This PR was automatically created by Snyk using the credentials of a real user.</p><br /><h3>Snyk has created this PR to upgrade org.apache.tinkerpop:tinkergraph-gremlin from 3.5.2 to 3.6.0.</h3>

:information_source: Keep your dependencies up-to-date. This makes it easier to fix existing vulnerabilities and to more quickly identify and fix newly disclosed vulnerabilities when they affect your project.
<hr/>

- The recommended version is **2 versions** ahead of your current version.
- The recommended version was released **23 days ago**, on 2022-04-05.

<hr/>

**Note:** *You are seeing this because you or someone else with access to this repository has authorized Snyk to open upgrade PRs.*

For more information:  

🧐 [View latest project report](https://app.snyk.io/org/svanteschubert/project/0138f2c5-41ad-48b7-8d5b-4f0ea350a540?utm_source&#x3D;github&amp;utm_medium&#x3D;referral&amp;page&#x3D;upgrade-pr)

🛠 [Adjust upgrade PR settings](https://app.snyk.io/org/svanteschubert/project/0138f2c5-41ad-48b7-8d5b-4f0ea350a540/settings/integration?utm_source&#x3D;github&amp;utm_medium&#x3D;referral&amp;page&#x3D;upgrade-pr)

🔕 [Ignore this dependency or unsubscribe from future upgrade PRs](https://app.snyk.io/org/svanteschubert/project/0138f2c5-41ad-48b7-8d5b-4f0ea350a540/settings/integration?pkg&#x3D;org.apache.tinkerpop:tinkergraph-gremlin&amp;utm_source&#x3D;github&amp;utm_medium&#x3D;referral&amp;page&#x3D;upgrade-pr#auto-dep-upgrades)

PR #170, opened 2022-04-29
Update Maven JavaDoc plugin due to: 'Error while generating Javadoc: …

svanteschubert:
…Unable to compute stale date: Input length = 1' and moving updated mavin plugin versions to root pom.xml

Triggered by a failed build due to a version update: https://github.com/tdf/odftoolkit/actions/runs/2243991608
I ran into a JavaDoc build problem with the latest Oracle JDK 17 build 2 (also with the latest build 3) and the latest Windows 10 updates (succeeded under Ubuntu Linux using OpenJDK 11).

This build problem could be fixed by updating to the latest Maven JavaDoc plugin.

Took the opportunity to update other Maven plugins and the Tinkerpop dependency that failed to build this morning.
Finally, mentioned the common maven plugin version as well in the root pom.xml and neglected the version in the child pom.xml. Sometimes additional configurations are necessary like for JavaDoc and JAR Maven plugin therefore a move was not possible, just a copy without configuration.

@all: Anything I might have overseen? Any further tweaks that come to your mind?
@mistmist Would you be so kind doing the review?

Build tested successfully using the above-mentioned platforms!

### DIFF
--- a/generator/pom.xml
+++ b/generator/pom.xml
@@ -43,7 +43,6 @@
             <plugins>
                 <plugin>
                     <artifactId>maven-compiler-plugin</artifactId>
-                    <version>3.8.1</version>
                     <configuration>
                         <!-- defined in the parent pom.xml -->
                         <source>${jdk.version}</source>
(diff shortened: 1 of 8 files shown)


svanteschubert:
Fixes #169

PR #179, opened 2022-09-01
getTextStyle has not public function visibility

Seb182:
Hi,

The visibility for the function is not indicated I think it is an oversight as all the others are public.


svanteschubert:
Thanks for the PR, Sébastien.
I was a bit distracted from the odf-codegen branch, but will merge your PR right after (Michael has reviewed & merged mine).  :-)

You are right, it seems to be an oversight you spotted.

svanteschubert (approved):
Thanks for your pull request (I have updated the master anyway), so I take over changes before merge of odf-codegen merge..

Perhaps meeting you at the LibreOffice conference next week :-)

Seb182:
Thank you, for the MR and also for your work on odftoolkit.

I won't be at the Milan conference :-( but I'll try to follow your talk remotely!

PR #185, opened 2022-11-01
Create CODE_OF_CONDUCT.md

bell-kevin:


mistmist (changes requested):
hi Kevin,

thank you for pointing out a Code of Conduct is missing in our repo.

we should definitely have one.

in fact there is already a Code of Conduct which applies to the ODF Toolkit project, it was just missing in the repository.

since this project belongs to The Document Foundation, its Code of Conduct applies.

it is available here: https://www.documentfoundation.org/foundation/code-of-conduct/

as it happens, it was just revised significantly a few weeks ago, and there are discussions underway to revise it further.

so i think it would be best to simply add a link, and not copy the content here where we'd have to update it (and most likely forget it).

i found this other TDF project already has a similar file:

https://github.com/tdf/libreoffice-metadata/blob/master/CODE_OF_CONDUCT.md

please update your PR to this and then we'll merge it - thanks!

mistmist (approved):
i think i found how to fix the rat problem, will merge this and fix it immediately

PR #198, opened 2022-12-19
clean up static Map/Set/etc to avoid races and leaks

mistmist:
Hopefully the OdfNamespace one won't grow to any measurable size, otherwise will have to be moved somewhere non-static.


svanteschubert (approved):
Thank you, Michael!

PR #221, opened 2023-05-23
Fixes #215: validator: build additional plain jar file

### LINKED ISSUE #215: Integrate ODF Validator using Maven
Hi!

I am having trouble integrating the ODF Validator through Maven in my Java project.

I can't find the proper import option as part of ```org.odftoolkit```.

I also had to add ```<type>pom</type>``` to the pom.xml dependency configuration. See issue: https://github.com/tdf/odftoolkit/issues/212

This is the code I am using, which in theory should work (I think):

```
import org.odftoolkit.odfdom.doc.OdfSpreadsheetDocument;
import org.odftoolkit.odfvalidator.ODFValidator;
import org.odftoolkit.odfvalidator.result.ValidationResult;

public class validate {
    // Validate OpenDocument Spreadsheets using ODF Validator
    public boolean Validate_ODFValidator(String filepath)  throws Exception {
        boolean valid = false;

        OdfSpreadsheetDocument spreadsheet = OdfSpreadsheetDocument.loadDocument(filepath);
        ODFValidator validator = new ODFValidator();
        ValidationResult result = validator.validate(spreadsheet);

        // Inform user and return boolean
        if (result.hasErrors()) {
            System.out.println("VALIDATE: The spreadsheet file is invalid.");
            System.out.println("VALIDATE: " + result.toString());
        } else {
            valid = true;
            System.out.println("VALIDATE: The spreadsheet file is valid.");
        }
        return valid;
    }
}
```

Can you help me?

mistmist:
Neither the .war file nor the -with-dependencies.jar file can be easily reused as a Maven dependency; let's try if this works better.

### DIFF
--- a/validator/pom.xml
+++ b/validator/pom.xml
@@ -101,6 +101,7 @@
                 <artifactId>maven-war-plugin</artifactId>
                 <version>3.3.2</version>
                 <configuration>
+                    <archiveClasses>true</archiveClasses>
                     <archive>
                         <index>true</index>
                         <manifest>


svanteschubert (approved):
As you described in the issue https://github.com/tdf/odftoolkit/issues/215 this Maven configuration should work! You might want to test it with a SNAPSHOT release: mvn deploy
as described in https://tdf.github.io/odftoolkit/release-guide.html

svanteschubert:
@mistmist It does not seem to work:
https://oss.sonatype.org/content/repositories/snapshots/org/odftoolkit/odfvalidator/0.12.0-SNAPSHOT/
but as it works for ODFDOM 
https://oss.sonatype.org/content/repositories/snapshots/org/odftoolkit/odfdom-java/0.12.0-SNAPSHOT/
the task is likely to copy/paste from it's pom.xml

svanteschubert:
The simple jar is already missing in the validator target folder therefore a local test seems sufficient!

PR #236, opened 2023-06-22
Improved navigation interface, added next() method

rsoika:
This is a smaller improvement of the `org.odftoolkit.odfdom.incubator.search.Navigation` Interface.  

Using this feature in my own project I recognized that the `TextNavigation` was not working correctly - especially in case the search pattern occur multiple times in one document or paragraph. 

The following example shows how to replace all matches of a regex with a text using the new implementation:

```java
	private void replaceODFTextFragment(OdfTextDocument doc, String pattern, String replace)
			throws InvalidNavigationException {
		try {			
			TextNavigation searchPattern = new TextNavigation(pattern, doc);
			while (searchPattern.hasNext()) {
				OdfElement element = searchPattern.next(); //.getCurrentItem().getElement();
				String text=element.getTextContent();
				logger.info(" Current Item Text="+text);
				element.setTextContent(replace);
			}
		} catch (Exception e) {
			e.printStackTrace();
		}
	}
```

The method `TextNavigation.next()` is new and a convenient method for `getCurrentItem().getElement();`

An important bugfix is in the method `setCurrentTextAndGetIndex`. This method runs in a outOfBounds Exception in case the pattern length was equals to the replacement text length. This is now also fixed.

I have tested the new code in my [own project](https://github.com/imixs/imixs-adapters/tree/master/imixs-adapters-odf) and it works now perfect - even with text in embedded table cells.

### DIFF
--- a/odfdom/src/main/java/org/odftoolkit/odfdom/incubator/search/Navigation.java
+++ b/odfdom/src/main/java/org/odftoolkit/odfdom/incubator/search/Navigation.java
@@ -50,7 +50,7 @@ public abstract class Navigation {
    *
    * @return the current Selection result
    */
-  public abstract Selection getCurrentItem();
+  public abstract Selection getSelection();
 
   /**
    * check if the element match the user defined condition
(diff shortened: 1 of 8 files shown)


svanteschubert (approved):
Thank you, Ralph, for this patch fixing the search functionality.
As it was broken earlier the renaming (API change) should not matter to me.

PR #269, opened 2023-12-20
Updating dependencies

svanteschubert:
Updating dependencies before release

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -29,7 +29,7 @@
     <parent>
         <groupId>org.apache</groupId>
         <artifactId>apache</artifactId>
-        <version>28</version>
+        <version>31</version>
         <relativePath/>
     </parent>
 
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -29,7 +29,7 @@
     <parent>
         <groupId>org.apache</groupId>
         <artifactId>apache</artifactId>
-        <version>28</version>
+        <version>31</version>
```
mistmist, line 32: stupid question: what is this "apache" thing?

mistmist (approved):

pom.xml
svanteschubert: > stupid question: what is this "apache" thing?

It is a parent pom with basic dependencies, which is the default for all Apache projects. I like their tooling such as the -Ppedantic build option searching for files without copy-right header! :-)
see https://maven.apache.org/pom/asf/ for details

PR #277, opened 2024-01-24
Fixes #276: set failOnNoGitDirectory property for git-commit-id-maven…

### LINKED ISSUE #276: Release file can't be built due to lack of .git
When checking out out git repository and building odftoolkit with `maven install` it could be built successfully.  However if I download the release file from https://github.com/tdf/odftoolkit/releases/download/v0.12.0/odftoolkit-0.12.0-src.zip it couldn't be built because lack of .git 

Screenshot-2024-01-24-13-21-19

mistmist:
…-plugin

Obviously release source zip files don't contain a .git directory, so avoid failing the build in this case.

### DIFF
--- a/odfdom/pom.xml
+++ b/odfdom/pom.xml
@@ -187,6 +187,7 @@
                     <excludeProperties>
                         <excludeProperty>git.*.user.*</excludeProperty>
                     </excludeProperties>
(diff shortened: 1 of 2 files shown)


svanteschubert (approved):
Thanks, Michael!

PR #310, opened 2024-07-18
Line-ending normalizsation for ODF JSON regression tests

svanteschubert:
JSON references - when cloned via Git on Windows receive sometimes a Windows line-ending (dependent on git configuration), which causes trouble (test errors - false positive) when comparing test results line-based.

The JSON created from the ODTs during regression tests is always created with Linux line endings.

Therefore, the line endings of reference test files are being changed from Windows to Linux to avoid this uncertainty.


svanteschubert:
Tested under Windows and Linux.

mistmist (approved):
note to self: replaceAll 1st param is a regex and "\r\n?" replaces both Mac and Win lineends

PR #314, opened 2024-07-29
Fixing ODF DOM element cloning

### LINKED ISSUE #313: Create a full copy of a spreadsheet in an ODF Spreadsheet document using the ODF Toolkit.
I need to create a full copy of a table sheet in an ODF Spreadsheet document using ODF Toolkit.
I'm trying to clone like this:

`ods.getContentRoot().appendChild(ods.getContentRoot().getChildNodes().item(1).cloneNode(true));`

The copy is created, but incorrectly. When opening the saved file with OpenOffice Calc - the cell formatting is not copied, when opening in Microsoft Excel - the cells are not filled in the copy. I use ODF Toolkit version 0.12.0.

svanteschubert:
Fixes #313 - OdfElement cloneNode omitted wrongly some ODF attribute for cloning


svanteschubert:
PS: How the problem/solution became apparent:
I have opened the content.xml of the output document and compared the two tables (original and cloned) and realized that the @office:value-type="float" was missing in the cloned table. 
Therefore in the various office applications the default type string was applied and the content was left aligned (as all strings) contrary to the numbers on the right side.

mistmist (approved):

PR #320, opened 2024-09-06
Fixes #319: validator: the generator element is in meta.xml

### LINKED ISSUE #319: Error when using "-g" option
If we use latest odfvalidator to check some files with `-g` option, it will show error message like this:
`file:///path/to/odffile/META-INF/manifest.xml: Error: Invalid root element: manifest:manifest`
Every document will get this result even if using vanilla LibreOffice to create a file that contains only "test" or some other word.

mistmist:
Apparently this was always wrong, and nobody used -g?


svanteschubert (approved):
Well, a manifest is indeed close to a meta.xml but not close enough ;-)
Thanks for the fix, Michael!  👍

PR #330, opened 2024-10-01
Use ConcurrentMap instead of synchronized for class cache

jungnitz:
I am currently developing an application that generates lots of documents in parallel. While profiling I noticed a bottleneck in `OdfXMLFactory#getOdfNodeClass`.

Since almost all entries of the Class cache are known after reading some documents, it should be possible to reduce locking and therefore increase parallelism. I firstly tried replacing the `synchronized`-Blocks with a RW-Lock, which resulted in slightly slower execution times. Using a ConcurrentHashMap however almost doubled the document throughput in my application.

Therefore I propose the changes in the commit.


svanteschubert (approved):
Great! Thank you very much, Niklas, for your contribution! Very much appreciated!
With kind regards from Berlin, Svante

PR #334, opened 2024-11-07
Add ODF1.4 support to validator.

fweng322:
1. Move ODF 1.4 schema files from LibreOffice master to here
2. Add version entries
3. Add command line arguments
4. Add test case file


mistmist (approved):
thanks, all looks plausible to me.

PR #338, opened 2024-11-21
Support for AES256-CBC encryption

### LINKED ISSUE #138: Access to encrypted ODS file fails
I have created an ODS file with LibreOffice Calc which I have protected with a password.
I opened this file as follows:
```
odfPackage = OdfPackage.loadPackage(document.getFile(), openPassword, null);
odfDocument = OdfDocument.loadDocument(odfPackage);
```

This still works. After that I try to get the MasterPage as follows:
```
odfDocument.getOfficeMasterStyles().getMasterPage("Default");
```

Unfortunately, I get this exception:
```
java.security.InvalidAlgorithmParameterException: Wrong IV length: must be 8 bytes long
	at java.base/com.sun.crypto.provider.CipherCore.init(CipherCore.java:525)
	at java.base/com.sun.crypto.provider.BlowfishCipher.engineInit(BlowfishCipher.java:222)
	at java.base/javax.crypto.Cipher.implInit(Cipher.java:843)
	at java.base/javax.crypto.Cipher.chooseProvider(Cipher.java:901)
	at java.base/javax.crypto.Cipher.init(Cipher.java:1433)
	at java.base/javax.crypto.Cipher.init(Cipher.java:1364)
	at org.odftoolkit.odfdom.pkg.OdfPackage.decryptData(OdfPackage.java:2052)
	at org.odftoolkit.odfdom.pkg.OdfPackage.getBytes(OdfPackage.java:1870)
	at org.odftoolkit.odfdom.pkg.OdfPackage.getInputStream(OdfPackage.java:2172)
	at org.odftoolkit.odfdom.pkg.OdfFileDom.initialize(OdfFileDom.java:219)
	at org.odftoolkit.odfdom.dom.OdfStylesDom.initialize(OdfStylesDom.java:79)
	at org.odftoolkit.odfdom.pkg.OdfFileDom.<init>(OdfFileDom.java:102)
	at org.odftoolkit.odfdom.dom.OdfStylesDom.<init>(OdfStylesDom.java:64)
	at org.odftoolkit.odfdom.dom.OdfSchemaDocument.getStylesDom(OdfSchemaDocument.java:436)
	at org.odftoolkit.odfdom.dom.OdfSchemaDocument.getOfficeMasterStyles(OdfSchemaDocument.java:536)
```

I am using the current code from the Master Branch.
During debugging, I became aware of the decryptData method in the OdfPackage.
There, in line 2027, the AlgorithmElement is read from the manifest.xml. It has the following content in my file:
```
<manifest:algorithm manifest:algorithm-name="http://www.w3.org/2001/04/xmlenc#aes256-cbc" manifest:initialisation-vector="qiWmy8seRpFOREHbrzhbag=="/>
```

After the Base64 decoding in line 2033, iv is then 16 bytes long.
Is it possible that in line 2048 we should use the algorithm from the manifest instead of the Blowfish? So in my case aes256-cbc?

Thanks,
Lars

lawern:
This pull request fixes the problem with password-protected LibreOffice files described in Issue #138.
It is now possible to open files with the old Blowfish encryption as well as with the AES256-CBC encryption.
Unfortunately, I had to add BouncyCastle because I couldn't solve the problem with the existing libraries.

resolves #138

### DIFF
--- a/odfdom/src/main/java/org/odftoolkit/odfdom/pkg/OdfPackage.java
+++ b/odfdom/src/main/java/org/odftoolkit/odfdom/pkg/OdfPackage.java
@@ -38,6 +38,7 @@ import java.io.OutputStream;
 import java.io.UnsupportedEncodingException;
 import java.lang.reflect.Constructor;
 import java.net.URI;
+import java.nio.charset.StandardCharsets;
 import java.security.InvalidKeyException;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
@@ -73,6 +74,10 @@ import javax.xml.transform.URIResolver;
 import org.apache.commons.compress.archivers.zip.ZipArchiveEntry;
 import org.apache.commons.compress.archivers.zip.ZipArchiveOutputStream;
 import org.apache.xerces.dom.DOMXSImplementationSourceImpl;
+import org.bouncycastle.crypto.PBEParametersGenerator;
+import org.bouncycastle.crypto.digests.SHA1Digest;
+import org.bouncycastle.crypto.generators.PKCS5S2ParametersGenerator;
+import org.bouncycastle.crypto.params.KeyParameter;
 import org.odftoolkit.odfdom.doc.OdfDocument;
 import org.odftoolkit.odfdom.doc.OdfDocument.OdfMediaType;
 import org.odftoolkit.odfdom.pkg.manifest.AlgorithmElement;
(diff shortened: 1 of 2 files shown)


odfdom/src/main/java/org/odftoolkit/odfdom/pkg/OdfPackage.java
```diff
@@ -2025,32 +2030,44 @@ private byte[] decryptData(
...
+
+      if (isAES) {
+        // AES-256-CBC
+        md = MessageDigest.getInstance("SHA-256");
+        passBytes = md.digest(passBytes);
+        PBEParametersGenerator generator = new PKCS5S2ParametersGenerator(new SHA1Digest());
+        generator.init(passBytes, salt, 100000);
+        KeyParameter keyParam = (KeyParameter) generator.generateDerivedParameters(256);
+        key = new SecretKeySpec(keyParam.getKey(), "AES");
+        cipher = Cipher.getInstance("AES/CBC/NoPadding");
```
mistmist, line 2050: i believe AES uses W3C padding? (whereas Blowfish didn't need padding)

hmm ... i find this page: https://docs.oracle.com/en/java/javase/11/docs/specs/security/standard-names.html

this line probably?

ISO10126Padding 	This padding for block ciphers is described in [5.2 Block Encryption Algorithms](https://www.w3.org/TR/xmlenc-core/#sec-Alg-Block) in the W3C “XML Encryption Syntax and Processing” document.

odfdom/src/main/java/org/odftoolkit/odfdom/pkg/OdfPackage.java
```diff
@@ -2025,32 +2030,44 @@ private byte[] decryptData(
...
+
+      SecretKeySpec key;
+      Cipher cipher;
+      MessageDigest md;
+
+      if (isAES) {
+        // AES-256-CBC
+        md = MessageDigest.getInstance("SHA-256");
+        passBytes = md.digest(passBytes);
+        PBEParametersGenerator generator = new PKCS5S2ParametersGenerator(new SHA1Digest());
```
mistmist, line 2046: this is just a different name for PBKDF2, right?

mistmist (approved):
let's merge this now, if it works.

i think this could be further improved by reading various attributes and not hard-coding so many things, such as these:

<manifest:encryption-data>
  manifest:checksum-type

<manifest:start-key-generation>
  manifest:start-key-generation-name

<manifest:key-derivation>
  manifest:iteration-count
  manifest:key-derivation-name

<manifest:algorithm>
   manifest:algorithm-name

odfdom/src/main/java/org/odftoolkit/odfdom/pkg/OdfPackage.java
lawern: It only worked for me without padding.
I got my info from this StackOverflow posting: [https://stackoverflow.com/questions/62001716/encrypt-a-file-inside-an-ods-archive/62005258?r=Saves_UserSavesList#62005258](https://stackoverflow.com/questions/62001716/encrypt-a-file-inside-an-ods-archive/62005258?r=Saves_UserSavesList#62005258)

odfdom/src/main/java/org/odftoolkit/odfdom/pkg/OdfPackage.java
```diff
@@ -2025,32 +2030,44 @@ private byte[] decryptData(
...
+        generator.init(passBytes, salt, 100000);
```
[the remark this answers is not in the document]
lawern: Yes, we could also read these values from the manifest, but I didn't want to change too much of your code.

odfdom/src/main/java/org/odftoolkit/odfdom/pkg/OdfPackage.java
mistmist: ah yes it would need to be with HmacSHA1 for ODF

odfdom/src/main/java/org/odftoolkit/odfdom/pkg/OdfPackage.java
mistmist: hmm but that also says to use ISO10126Padding?

Issue #353, opened 2025-05-09
logging should be consistent and no logging implementations should be added as a transient dependency


xzel23:
Following problems:

- slf4j-simple is added as a transient dependency. A library project should never add a dependency on a logger implementation as we don't know what implementation that uses ODF Toolkit will use.

- a dependency on slf4j-api is added with the comment that Apache Jena needs it. If a dependency depends on a library, it is the responsibility of the dependency to declare it properly. If a logger implementation is needed for running tests, it should be included with "test" scope, and only one such implementation should be present. If different logging facades are used by dependencies, the corresponding bridging modules should be included with test scope.

- ODF Toolkit itself mainly uses JUL logging with one class using SLF4J instead. All logging done by ODF Toolkit should use the same API.

Also, the project should consider whether to use one of the two common logging APIs instead of calling JUL directly as that causes performance issues when an application uses a different logging implementation. IMHO either log4j-api or slf4j-api should be used instead of calling JUL directly (and my preference would be log4j-api since bridging from slf4j to log4j is less performant than the other way around.

I'll fix the inconsistent logging calls and the dependency issues and if the project agrees on using either log4j-api or slf4j-api, I can do the necessary changes later.

xzel23:
The logging PR was applied so we can change this for now. I still would like to replace JUL-logging with Log4J-API, but that should be done after the next release in a separate ticket.

PR #354, opened 2025-05-09
Issues/353 inconsistent logging

xzel23:
This fixes the usage of SLF4J in the single source file that uses SLF4J and removes the dependencies on SLF4J. If any dependency requires the SLF4J API, it is their responsibility to declare it as a transient dependency.

### DIFF
--- a/odfdom/pom.xml
+++ b/odfdom/pom.xml
@@ -89,14 +89,6 @@
             <artifactId>json</artifactId>
         </dependency>
         <dependency>
-            <groupId>org.slf4j</groupId>
-            <artifactId>slf4j-api</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>org.slf4j</groupId>
-            <artifactId>slf4j-simple</artifactId>
-        </dependency>
-        <dependency>
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-compress</artifactId>
         </dependency>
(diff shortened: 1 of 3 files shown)


xzel23:
Fixes part of #353

svanteschubert:
@xzel23 I am not actively using SLF4J, it was once necessary due to the Apache Jena dependency, and your patch works as you suggested, but as only tweak, we might want to add/keep the SLF4J API dependency as an advantage. At least this is how I read the manual:
image
The text in blue seems to indicate that keeping the explicit API dependency is of advantage. 

Any other suggestions? Anyone?

xzel23:
@svanteschubert Hi Svante, that text only applies to applications, library projects and frameworks (see the paragraph just above:

> Given Maven's transitive dependency rules, for "regular" projects (**not libraries or frameworks**) declaring logging dependencies can be accomplished with a single dependency declaration.

The paragraph you mentioned is about simple programs where you want a simple logging implementation and include slf4j-simple. It also pulls in slf4j-api. But due to maven's somewhat obscure dependency resolution rules, you can not be sure what exact version will be pulled in if a dependency also declares (another version of) slf4j-api as a transitive dependency. That is solved by declaring direct dependencies on both slf4j-simple and slf4j-api.

I just started looking at ODF Toolkit, so I don't know yet if it also builds applications. If so, you will see a warning that no Logging implementation for SLF4J could be found and a suggestion to add slf4j-simple. But that should only be done for application code, never for libraries.

The problem with libraries is that you as a library developer don't know what logging implementation the application someone else creates will be using, and pulling in a logging implementation will lead to the application using several different logging applications that will write to different log files, use different output formats, and have to be configured separately.

The library should simply use a logging facade (log4j-api or slf4j-api), so that the application developer is free to use whatever logging implementation they want or have to use. Logging calls by libraries are then rerouted to that implementation by adding a dependency on the appropriate brigde module. Like that all logging ends up in a single place, uses a single configuration and is done using the same format.

slf4j-simple is usually used as a minimal logger for unit tests (with the dependency declared to use the "test" scope) or very basic applications.

svanteschubert (approved):
Thanks for your patch and your explanations!

PR #356, opened 2025-05-09
fix #355: remove unnecessary calls to toString()

### LINKED ISSUE #355: lots of unnecessary calls to toString()
The code contains lots of unnecessary calls to toString(). This is not just cosmetic: when a logging call is made, the arguments will be evaluated regardless whether logging at the current level is enabled or not, which causes unnecessary overhead, both for the cpu and the garbage collector.

Patch incoming.

xzel23:


svanteschubert (approved):
Thanks for the patch!

PR #357, opened 2025-05-10
cleanup logging

xzel23:
the next PR for #353 

- loggers should be declared private static final
- loggers should not be initialized with a foreign class


svanteschubert (approved):
Thank you very much for taking the time to fix this! I do love consistency! :-)

Issue #360, opened 2025-05-10
re-enable disabled unit tests and make sure they pass


xzel23:
Title says it all. I'll create a series of patches. But this will take some time.

PR #363, opened 2025-05-12
Fix/360 re enable disabled unit tests

xzel23:
This is the first part of fixes for #360. It is about TableCellTest. The whole test class was ignored. After enabling it again, many tests failed.

Many of the tests work by loading a workbook, making changes, saving the workbook again and then reloading it to assert that the correct results are read from the updated workbook. The paths were messed up and some tests read the original workbook instead of the updated one when reloading, others  used the wrong path and a FileNotFoundException occurred.

In the first step, I removed the @Ignore on the test class and made sure the correct files are read. Some tests still fail which indicates that there are bugs in ODF Toolkit. I added `@Ignore` annotations and a FIXME comment to these.

As I am still quite new to this project, someone with more knowledge about the internal workings might look at some of the failing tests, for example formats are written out but come back differently after reloading.

But I think this PR is already an improvement because now we have at least 24 working tests where before, there were 0.

### DIFF
--- a/odfdom/src/test/java/org/odftoolkit/odfdom/doc/table/TableCellTest.java
+++ b/odfdom/src/test/java/org/odftoolkit/odfdom/doc/table/TableCellTest.java
@@ -55,19 +55,16 @@ public class TableCellTest {
   static final String SAMPLE_TEXT = "TestTextTable";
   OdfSpreadsheetDocument odsdoc, odsstyle;
   OdfTextDocument odtdoc;
-  OdfTable odsTable, odtTable;
 
   @Before
   public void setUp() {
     try {
       odsdoc =
-          (OdfSpreadsheetDocument)
-              OdfSpreadsheetDocument.loadDocument(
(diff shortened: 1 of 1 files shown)


xzel23:
Wait, there's more coming for this class.

xzel23:
PR is updated.

I have removed most of the exception handling code and instead declared exceptions on the test methods. The old code caught the exception, logged it, and failed the test with the message obtained from the exception. That's pretty much the same that JUnit does with exceptions thrown from test code, so let JUnit do this to keep the code clear.

I also replaced the manual checking for thrown exceptions with assertThrows() which does the same in a more concise way.

xzel23:
@svanteschubert I think this is ready to merge now. There still are 5 failing tests, I have added comments about the failures. I think these are bugs in the library. I'll move on to the next test class now. It would be great if someone could look at these failures in the meantime.

svanteschubert (approved):
@xzel23 Thanks for the patch!
I started looking into the broken tests and will provide a patch tomorrow

PR #366, opened 2025-05-12
Fix/365 replace depracated constructor usage

### LINKED ISSUE #365: Primitive constructors are deprecated for removal and should be replaced
See title.

Patch incoming. Also, in most (all?) places, it would be better to directly use the primitive types directly, but that's something for later.

xzel23:
fixes #365


xzel23:
Please skip this one. Changes are also included in #367

Issue #367, opened 2025-05-12
cleanup the code


xzel23:
I will collect several PRs here since there are many places where the code can be simplified and putting all in a single PR will be hard to review.

1. ~~resource handling: use try-with resources to reduce complexity and handle corner cases correctly; don't re-implement JDK functionality~~
2. ~~remove unnecessary explicit boxing and unboxing~~
3. ~~simplify operations on collections, for example use Map.putIfAbsent()/computeIfAbsent() and others~~
4. ~~collapse Exceptions with the same catch clauses~~
5. ~~replace StringBuffer with StringBuilder~~

Please keep this one open for the time being, I will provide iterative PRs.

mistmist:
thanks Axel for your many efforts!
the PRs are appreciated, but i think a github issue for generic code cleanups isn't required... just put a short description of the cleanup in the commit message, file as many PRs as you like... it's always "open season" for self-contained code cleanups, and issues should be filed for large-scale refactorings that need discussion, and bugs.

xzel23:
There are still some places left in the code, but I will close this now.

PR #369, opened 2025-05-13
cleanup string operations

xzel23:
fixes 5. of #367


svanteschubert (approved):
What a nice spring cleaning - reviewing this makes me realise that we started before JDK 1.5 ;-)
Thanks again for the patch, Axel! 👍

Issue #371, opened 2025-05-13
Fixing the roundtrip of Cell FormatCode (implementation-defined) for Time/Calendar type


svanteschubert:
A FormatCode is being used in ODF for Cells, which is implementation-defined by the ODF applications:
![image](https://github.com/user-attachments/assets/490b96b7-5b2b-40f3-83cc-fbf8be621fea)

For instance, 

1. **Excel** defines it here: https://support.microsoft.com/en-us/office/review-guidelines-for-customizing-a-number-format-c0a1d1fa-d3f4-4018-96b7-9c9354dd99f5 and their example of a FormatCode 
"[Blue]#,##0.00_);[Red](#,##0.00);0.00;"sales "@" 
will result in a positive integer being blue 1234.5 will be displayed as 1.234.50 in blue, while the same as a negative number with brackets in red.
Any string data will receive a sales prefix.
I was able to add this string in LibreOffice 24.8.6.2 in the GUI for custom formatting of a cell, and it worked!

2. **LibreOffice** defines it here: https://help.libreoffice.org/24.8/en-GB/text/shared/01/05020301.html?&DbPAR=CALC&System=WIN

**Reading the above documentation, there is no difference between yyyy and YYYY or dd and DD - all calendar characters have been provided in capital letters in the documentation of LibreOffice, which we might want to align by default as we are part of the same provider ( TDF)!
Please correct me if I misinterpreted this.**

In our [test class TableCellTest.java](https://github.com/tdf/odftoolkit/blob/master/odfdom/src/test/java/org/odftoolkit/odfdom/doc/table/TableCellTest.java), the test testSetDefaultCellStyle() was disabled as it was not being round-tripped correctly.

**In addition, the implementation of DD and dd is currently different in ODFDOM, while DD is the day of the month, dd is the day of the year, which should be fixed to the day of the month!**

Please consider that the FormatCode has been changed, and the latest version of LibreOffice should be used for updating/validating the test documents.

svanteschubert:
The format codes are being transformed to ODF XML nodes, and when reading them, we might want to return all calendar symbols as capital letters and accept both kinds.

xzel23:
I don't think it is even possible to always accept both forms and knowing what is meant.

For some fields, upper- and lowercase symbols are used but with different meanings, like 'h' and 'H', 'd' and 'D', 'y' and 'Y'. While for 'm' and 'M' this can be derived from context, the other cases cannot really be distinguished.

[...]

Issue #370, opened 2025-05-13
Fixing the roundtrip of Cell FormatCode (implementation-defined) for Currency


svanteschubert:
A FormatCode is being used in ODF for Cells, which is implementation-defined by the ODF applications:
![image](https://github.com/user-attachments/assets/490b96b7-5b2b-40f3-83cc-fbf8be621fea)

For instance, 

1. **Excel** defines it here: https://support.microsoft.com/en-us/office/review-guidelines-for-customizing-a-number-format-c0a1d1fa-d3f4-4018-96b7-9c9354dd99f5 and their example of a FormatCode 
"[Blue]#,##0.00_);[Red](#,##0.00);0.00;"sales "@" 
will result in a positive integer being blue 1234.5 will be displayed as 1.234.50 in blue, while the same as a negative number with brackets in red.
Any string data will receive a sales prefix.
I was able to add this string in LibreOffice 24.8.6.2 in the GUI for custom formatting of a cell, and it worked!

2. **LibreOffice** defines it here: https://help.libreoffice.org/24.8/en-GB/text/shared/01/05020301.html?&DbPAR=CALC&System=WIN

Reading the above documentation, the # is an optional integer, which will be omitted when not present in the number.
From my understanding, the example #0.0 does not make sense in the text, only #.0, where the number .9 will remain without a leading 0. 
Please correct me if I misinterpreted this. 

In our [test class TableCellTest.java](https://github.com/tdf/odftoolkit/blob/master/odfdom/src/test/java/org/odftoolkit/odfdom/doc/table/TableCellTest.java), the test testGetSetCurrencyFormat() was disabled as it was not being round-tripped correctly.

Please consider that the FormatCode has been changed, and the latest version of LibreOffice should be used for updating/validating the test documents.

Issue #364, opened 2025-05-12
Loading an invalid ODF without version attribute in manifest:manifest should add it


svanteschubert:
The ODS test 
https://github.com/tdf/odftoolkit/blob/master/odfdom/src/test/java/org/odftoolkit/odfdom/doc/table/TableCellTest.java
loads the (invalid) test document [TestSpreadsheetStyleTable.ods](https://github.com/tdf/odftoolkit/blob/master/odfdom/src/test/resources/test-input/TestSpreadsheetStyleTable.ods), changes it, and saves it back, but there is still a version attribute missing in the manifest:

![Image](https://github.com/user-attachments/assets/533b993f-afb8-43d7-a1af-3795871e185c)

tested with https://odfvalidator.org/

Such a missing version attribute should be automatically added when it does not exist.
This might happen on the DOM level, and I will take this issue to generate this feature.

Issue #229, opened 2023-05-27
Incorrect behavior of OdfElement.getTextContent()


ttaomae:
While investigating #220, I came across `OdfTableCell.getOdfElement().getTextContent()`, which is defined by [`org.w3c.dom.Node.getTextContent()`](https://docs.oracle.com/en/java/javase/11/docs/api/java.xml/org/w3c/dom/Node.html#getTextContent()) and implemented [in `OdfElement`](https://github.com/tdf/odftoolkit/blob/v0.11.0/odfdom/src/main/java/org/odftoolkit/odfdom/pkg/OdfElement.java#L2638). My interpretation of the `Node.getTextContent()` documentation is that it should recursively retrieve the text contents of all child nodes and concatenate them together. However, the `OdfElement` implementation only looks at the children, rather than all descendants.

I think it is also worth noting that `OdfWhitespaceParser.getText(Node)` appears to do what `OdfElement.getTextContent()` is supposed to do. Or at least something very similar.

mistmist:
i don't know the ODFDOM code well, but as a general rule about ODF: inside of a paragraph (text:p/text:h) and its child elements, there is character content which is part of the paragraph text (most elements), and there is character content which is part of some nested object and not part of the paragraph text.

there are a bunch of elements for drawing shapes, draw:frame etc. and also office:annotation, that should comprise most of the latter case - these should probably not be included in any API that promises some sort of text content of "the paragraph" or "the table cell".

the criteria for what is considered paragraph text are in section "6.1.2 White Space Characters" of ODF 1.3 part 3, and i think the text of a table cell should simply be the concatenation of its paragraphs.

svanteschubert:
You are absolutely right, Todd.

OdfElement has the base functionality to concatenate the text content:
https://github.com/tdf/odftoolkit/blob/master/odfdom/src/main/java/org/odftoolkit/odfdom/pkg/OdfElement.java#L2633

but the every text node containing element like OdfTextSpan should override this method and define its specific behavior.
By this method implementation, the specific behavior - Michael has pointed out it his previous comment - would be reflected.

I am uncertain atm, if we have a way in the ODF Specification to differentiate this behavior, if not it would be worth to add it.

Taking this over issue over!
Will aim to generate this as part of the upcoming release candidate.

Thanks for pointing out!

svanteschubert:
The ODF specification explains in the bullet list of 6.1.1 all elements containing text that should not be considered, see
[...]

PR #372, opened 2025-05-13
Enabling FormatCodeTests

svanteschubert:
Enabling FormatCodeTests and created follow-up issues for the remaining problems, i.e. #370, #371 and by updating a test file (using different dates for different formatting to be certain to have the correct cell) and using pretty printing XML more assertions had to be disabled due to #229

What has been changed:

1. First of all, I realised that the output was invalid, but so was the input from IBM Lotus Symphony, released October 2010, which omitted the version attribute in the manifest, created #364 to fix invalid files (at least in this case) by adding a version.
<br/>then I had to (re)learn this formatting language for spreadsheet cells, which seems to be implementation-dependent:
image
see https://docs.oasis-open.org/office/OpenDocument/v1.4/OpenDocument-v1.4-part4-formula.html#TEXT
But the ODF implementations describe this formatting language:
- **EXCEL:** https://support.microsoft.com/en-us/office/review-guidelines-for-customizing-a-number-format-c0a1d1fa-d3f4-4018-96b7-9c9354dd99f5
- **LibreOffice**: https://help.libreoffice.org/24.8/en-GB/text/shared/01/05020301.html?&DbPAR=CALC&System=WIN
I still could paste the Excel String "[Blue]#,##0.00_);[Red](#,##0.00);0.00;"sales "@" in the latest LibreOffice as custom formatting and it worked, so they are not so far away.
<br/>After reading, I realised that the example #0.0 does not make any sense at all, so I exchanged it with #.0

2. Second, the formats that I received from LibreOffice differ from the expected ones from the tests, so I simply resaved the document and took the date formatting from the file, which looked good, and changed the dates to be able to prove which cell was taken by the content.

3. Third, I made some Nullpointer sanity checks, but I avoided changing the API, like sometimes an InvalidArgumentException is thrown, when the cell is not of the time type, but if not set, Null is returned.


odfdom/src/main/java/org/odftoolkit/odfdom/doc/table/OdfTableCell.java
```diff
@@ -507,13 +507,13 @@ public TableTableCellElementBase getOdfElement() {
    *     is not "currency".
    */
   public String getCurrencyCode() {
-    if (mCellElement
-        .getOfficeValueTypeAttribute()
-        .equals(OfficeValueTypeAttribute.Value.CURRENCY.toString())) {
-      return mCellElement.getOfficeCurrencyAttribute();
-    } else {
-      throw new IllegalArgumentException();
+    String t = mCellElement.getOfficeValueTypeAttribute();
```
xzel23, line 510: How about simply doing this slightly shorter and a message that also shows the unexpected value?

```
if (!Objects.equals(t, OfficeValueTypeAttribute.Value.CURRENCY.toString()) {
    LOG.log(Level.WARNING,"@office:value-type attribute expected to be of type CURRENCY but was {0}", t);
}

```

xzel23:
Ah, I thought this was only about the format strings. I have changed the code to not use the obsolete pre-Java 8 time classes and the not threadsafe SimpleDateFormatter. The java.time classes already use ISO 8601 out of the box when parsing or calling toString(), so this all gets a lot easier. When using custom format strings, the DateTimeFormatter accepts the same formats as SimpleDateFormatter (plus some added fields), but it is more picky and will refuse incorrect formats that SimpleDateFormatter will silently process.

odfdom/src/main/java/org/odftoolkit/odfdom/doc/table/OdfTableCell.java
```diff
@@ -919,24 +919,28 @@ public String getStringValue() {
...
+   * Get the cell value as {@link java.util.Calendar java.util.Calendar}
    *
    * <p>Throw exception if the cell type is not "time".
    *
-   * @return the Calendar value of cell
+   * @return the Calendar value of cell, which might be NULL if no value was set
    * @throws IllegalArgumentException an IllegalArgumentException will be thrown if the cell type is
    *     not time.
    */
   public Calendar getTimeValue() {
```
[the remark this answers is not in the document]
svanteschubert: Perhaps we should do a release this or end of summer, that is only adding new updated APIs, but keeps the old pre-Java 8 APIs to keep the pain of the existing users low.
My suggestion would be to do a version 1.0.0 with quite a few APIs that we currently have, perhaps deprecating some.
The removal of the "incubator" package name will already be painful enough - we should have used annotations like Google Guava @Beta annotation.

odfdom/src/main/java/org/odftoolkit/odfdom/doc/table/OdfTableCell.java
svanteschubert: Cool! I was not aware of the null-safe _Objects.equals(..)_ will surely overtake it, and I do like your improvement of giving out what the incorrect value is! Thanks!😎👍

odfdom/src/main/java/org/odftoolkit/odfdom/doc/table/OdfTableCell.java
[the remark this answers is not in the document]
xzel23: Yes, that's exactly what I did. All the old methods should work without observable change for users. They internally delegate to the new ones. So there really shouldn't be any pain. The old methods are marked as deprecated to encourage users to use the new ones.

If any method using the pre-Java 8 classes does not work with the PR applied exactly as it did before, that's a bug and I will fix it. ;-)

xzel23:
The problem is the 'M'. Both SimpleDateFormat (pre-Java 8) and DateTimeFormatter (java.time) interpret this as 'Month of year', and there is no month in a time (without a date). For minutes, 'm' has to be used for both formatting classes. Also, 'S' is for milliseconds, so `HH:MM:SS` should really be `HH:mm:ss`. So why is it not all lowercase? Because `h` is for the hour (1-12) in am/pm time.

I'll have a look at this PR later.

odfdom/src/test/java/org/odftoolkit/odfdom/doc/table/TableCellTest.java
```diff
@@ -270,12 +272,11 @@ public void testSetGetFormat() throws Exception {
     OdfTableCell cell = tablerow.getCellByIndex(3);
     Calendar currenttime = Calendar.getInstance();
     cell.setDateValue(currenttime);
-    cell.setFormatString("yyyy-MM-dd");
+    cell.setFormatString("YYYY-MM-dd");
```
xzel23, line 275: use `yyyy` here, `YYYY`is the week-based year, which will give wrong results in the first week of january if the week starts in the year before.

odfdom/src/test/java/org/odftoolkit/odfdom/doc/table/TableCellTest.java
```diff
@@ -313,30 +314,30 @@ public void testSetGetFormat() throws Exception {
     for (int i = 1; i <= 10; i++) {
       cell = tbl.getCellByPosition("A" + i);
       cell.setDateValue(Calendar.getInstance());
-      cell.setFormatString("yyyy.MM.dd");
+      cell.setFormatString("YYYY.MM.dd");
```
xzel23, line 317: small `yyyy`is correct here, see above.

odfdom/src/test/java/org/odftoolkit/odfdom/doc/table/TableCellTest.java
```diff
@@ -313,30 +314,30 @@ public void testSetGetFormat() throws Exception {
...
     }
     cell = tbl.getCellByPosition("A11");
     cell.setFormula("=max(A1:A10)");
     // contains 'y' 'M' 'd' should be adapted as date.
-    cell.setFormatString("yyyy.MM.dd");
+    cell.setFormatString("YYYY.MM.dd");
```
xzel23, line 322: `yyyy`

odfdom/src/test/java/org/odftoolkit/odfdom/doc/table/TableCellTest.java
```diff
@@ -313,30 +314,30 @@ public void testSetGetFormat() throws Exception {
...
       cell = tbl.getCellByPosition("A" + i);
       cell.setTimeValue(Calendar.getInstance());
-      cell.setFormatString("yyyy.MM.dd HH:mm:ss");
+      cell.setFormatString("YYYY.MM.dd HH:MM:SS");
     }
     cell = tbl.getCellByPosition("A11");
     cell.setFormula("=max(A1:A10)");
     // contains 'H' 'm' 's' should be adapted as time.
-    cell.setFormatString("yyyy.MM.dd HH:mm:ss");
+    cell.setFormatString("YYYY.MM.dd HH:MM:SS");
```
xzel23, line 335: `    cell.setFormatString("yyyy.MM.dd HH:mm:ss");`

odfdom/src/test/java/org/odftoolkit/odfdom/doc/table/TableCellTest.java
```diff
@@ -313,30 +314,30 @@ public void testSetGetFormat() throws Exception {
...
     Assert.assertEquals("time", cell.getValueType());
     cell = tbl.getCellByPosition("A12");
     cell.setFormula("=max(A1:A10)");
     // contains 'H' 'm' 's' should be adapted as time.
-    cell.setFormatString("HH:mm:ss");
+    cell.setFormatString("HH:MM:SS");
```
xzel23, line 340: `cell.setFormatString("HH:mm:ss");`

odfdom/src/test/java/org/odftoolkit/odfdom/doc/table/TableCellTest.java
```diff
@@ -234,34 +235,35 @@ public void testGetSetTextValue() throws Exception {
...
     Assert.assertEquals(
         "300" + (new DecimalFormatSymbols()).getDecimalSeparator() + "0", displayvalue);
     String format = fcell.getFormatString();
-    Assert.assertEquals("#0.0", format);
+    Assert.assertEquals("#.0", format);
 
     OdfTableCell dcell = table.getCellByPosition(3, 2);
     format = dcell.getFormatString();
-    Assert.assertEquals("MMM d, yy", format);
+    Assert.assertEquals("D. MMM YY", format);
```
xzel23, line 256: You used the german format with a day-of-year and week-based-year here. I think this should be `d. MMM yy`. The old code used the american format. Which one to use depends on the input spreadsheet and maybe the locale setting, but I have to check the spreadsheet first.

odfdom/src/test/java/org/odftoolkit/odfdom/doc/table/TableCellTest.java
```diff
@@ -234,34 +235,35 @@ public void testGetSetTextValue() throws Exception {
...
 
-    dcell.setFormatString("yyyy-MM-dd");
+    dcell.setFormatString("YYYY-MM-dd");
```
xzel23, line 258: old one is correct

odfdom/src/test/java/org/odftoolkit/odfdom/doc/table/TableCellTest.java
```diff
@@ -1025,9 +1028,9 @@ public void testSetDefaultCellStyle() throws Exception {
     contentAutoStyles = contentDom.getOrCreateAutomaticStyles();
 
     OdfNumberDateStyle dateStyle =
-        new OdfNumberDateStyle(contentDom, "yyyy-MM-dd", "numberDateStyle", null);
+        new OdfNumberDateStyle(contentDom, "YYYY-MM-dd", "numberDateStyle", null);
```
xzel23, line 1031: `yyyy`

odfdom/src/test/java/org/odftoolkit/odfdom/doc/table/TableCellTest.java
```diff
@@ -1050,7 +1053,8 @@ public void testSetDefaultCellStyle() throws Exception {
     OdfTableCell aCell = column.getCellByIndex(0);
     aCell.setValueType("date");
     String format = aCell.getFormatString();
-    Assert.assertEquals("yyyy-MM-dd", format);
+    // due to https://github.com/tdf/odftoolkit/issues/371
+    Assert.assertEquals("YYYY-MM-dd", format);
```
xzel23, line 1057: `Assert.assertEquals("yyyy-MM-dd", format);`

odfdom/src/main/java/org/odftoolkit/odfdom/doc/table/OdfTableCell.java
```diff
@@ -1126,6 +1127,10 @@ public LocalTime getLocalTimeValue() {
     if (getTypeAttr() == OfficeValueTypeAttribute.Value.TIME) {
       String timeStr = mCellElement.getOfficeTimeValueAttribute();
       try {
+        if (timeStr == null) {
```
xzel23, line 1130: I don't think this is an improvement: the old code would throw an exception, the new code more or less silently returns null. Further down, this may lead to incorrect results. Better throw an exception. If the user wants to handle this leniently, they should catch the exception.
If on the other hand empty cells with time value type should be allowed, we should return an Optional instead. Are empty time cells allowed in ODF? I don't know.

odfdom/src/test/java/org/odftoolkit/odfdom/doc/table/TableTest.java
```diff
@@ -885,7 +885,8 @@ public void testGetCellByPosition() throws Exception {
     table = mOdsDoc.getTableByName("Sheet1");
     cell = table.getCellByPosition("C1");
     Assert.assertNotNull(cell);
-    Assert.assertEquals("Currency", cell.getStringValue());
+    // FixMe: Testfile now with pretty printing in XML https://github.com/tdf/odftoolkit/issues/229
```
xzel23, line 888: I don't understand this comment. Does the pretty-printing change the string value?

odfdom/src/main/java/org/odftoolkit/odfdom/doc/table/OdfTableCell.java
```diff
@@ -613,7 +612,7 @@ public void setCurrencyCode(String currency) {
     if (currency == null) {
```
xzel23, line 612: Not really related to this particular PR, but I see IllegalArgumentExceptions thrown all over the place for null arguments. There are points for and against using IllegalArgumentException instead of NullPointerException. When using NPE, we could simply use `Objects.requireNonNull(currency)` or `Objects.requireNonNull(currency, "Currency code should not be null")`.

If we use IAE instead, we should add such a utility method in the codebase so this becomes:

```
    public void setCurrencyCode(String currency) {
        checkParameterNonNull(currency, "Currency code must not be nulll");
        ...
    }
```

Or use JSpecify anntotations (that makes the user's IDE, Spotbugs and other tools emit warnings) and possibly use a tool such as cabe to automatically have those checks injected during compilation based on the annotations (note: as the author of Cabe, I'm naturally biased here ).

This would simply look like this:
```
@NullMarked
class OdfTableCell {
    ...
    // as the whole class is @Nullmarked, currency must not be null
    // the user's IDE (if correctly configured) will warn about a (possibly) null value
    public void setCurrencyCode(String currency) { 
        ...
    }
}
```

Only where null is allowed, the parameter has to be annotated with `@Nullable`.

odfdom/src/test/java/org/odftoolkit/odfdom/doc/table/TableCellTest.java
```diff
@@ -270,12 +272,11 @@ public void testSetGetFormat() throws Exception {
...
     Calendar currenttime = Calendar.getInstance();
     cell.setDateValue(currenttime);
-    cell.setFormatString("yyyy-MM-dd");
+    cell.setFormatString("YYYY-MM-dd");
     tablerow = table.getRowByIndex(7);
     cell = tablerow.getCellByIndex(3);
     cell.setTimeValue(currenttime);
-    cell.setFormatString("HH:mm:ss");
-
+    cell.setFormatString("HH:MM:SS");
```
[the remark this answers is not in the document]
xzel23: The LibreOffice Strings are all locale dependent. If we want to use the LibreOffice strings directly, a translation method has to be implemented with lookups for all supported locales. I think, internally, the ISO 8601 format should be used and a translation method if the library user wants to use the LibreOffice format strings (which as I understand it are not part of the ODF spec).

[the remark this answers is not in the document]
xzel23: I think we should maybe add additional `String toLibreOfficeFormat(String isoFormat, Locale locale)` and `String parseLibreOfficeFormat(String loFormat, Locale locale)` as convenience methods for users. To have localized meanings when calling `setFormatString()` is a nightmare when you want to develop an application that should be able to run on different locales.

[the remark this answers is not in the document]
mistmist: > The LibreOffice Strings are all locale dependent.

AFAIK LibreOffice displays locale dependent stuff in the UI but always stores it as en-US (i.e. effectively locale-independent) in ODF.

(in contrast, i've had to solve problems related to importing locale-dependent text fields from DOCX....)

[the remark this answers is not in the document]
xzel23: Yes, so the formatting code needs some updates. I think it would be a good thing to support the (english) LibreOffice format strings as well as the ISO format, using two different methods. I see if I can put together a two-way-converter building on what is already there and make a suggestion in code (PR).
That way, compatibility for existing users is maintained and we also have a more "standard" (from the Java perspective) way of dealing with formats. Ideally, this will cover all formats, but I'll start with dates and times.

svanteschubert:
Let's summarise a bit on the date patterns we have been discussing.

We are an ODF library, our ground truth is the ODF specification. 
Unfortunately, the Date patterns are not part of the ODF specification - reusing simply the [ISO standard 8601](https://en.wikipedia.org/wiki/ISO_8601) standard - but every ODF application can use its date patterns, see :
image
see https://docs.oasis-open.org/office/OpenDocument/v1.4/OpenDocument-v1.4-part4-formula.html#TEXT

These patterns are named in the ODF specification **FormatCode**, which is a sequence of characters with an implementation-defined meaning.
Unfortunately, the ISO 8601 standard is usually behind a pay wall, but there is a [PDF draft of working group 5 available from 2016](https://www.loc.gov/standards/datetime/iso-tc154-wg5_n0038_iso_wd_8601-1_2016-02-16.pdf). 
But an implementation of this standard might be - hard to prove, not having the spec - the Java implementation available since JDK 8: See https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html#patterns

Nevertheless, in the file formats - within the XML - only the "FormatCode" of an ODF application will be found and must be translated back and forth from their implementation-defined to the JDK class.

In this issue, we might start making the existing code, which relies on IBM Lotus Symphony (based on OpenOffice.org), run with LibreOffice:
The FormatCode of LibreOffice is defined in its online help:
https://help.libreoffice.org/24.8/en-GB/text/shared/01/05020301.html?&DbPAR=CALC&System=WIN

When all @Ignore are being removed from 
https://github.com/tdf/odftoolkit/blob/master/odfdom/src/test/java/org/odftoolkit/odfdom/doc/table/TableCellTest.java
The problems become apparent, in addition to the text normalisation #229, which I will focus on now.
@xzel23 Would you like to give this format code fix a try, as it was related to the time/date refactorings you did earlier?

If anyone likes to provide MS Office ODF FormatCode support, any PR would be welcome! :-)

xzel23:
Hi @svanteschubert, yes, I wanted to do that anyway. My suggestion is to make the get- and set-methods accept the LibreOffice format strings, but also add alternative methods, maybe setIsoDateFormat() or setJavaDateFormat(), that accept the Java SimpleDateFormatter/DateTimeFormatter ISO compliant codes, because that's what Java developers are used to and many have in their code base.

PS: The symbols to be used in format strings are defined in the first part of the standard ("ISO 8601-1:2019(en)"), under "3.2.4   Symbols used in place of digits or signs". While you need to pay (IMHO too much) money to obtain a copy of the standard, that section is available free to read when you click the "Read Sample" below the rendered page on the left button of the [ISO 8601 page](https://www.iso.org/standard/70907.html).

PR #200, opened 2022-12-22
Release candidate: ODF TOOLKIT 0.12.0

svanteschubert:
For the release candidate of ODF TOOLKIT 0.12.0 I have done so far the following:

1. Adjusted dependencies of ODF 1.2 to ODF 1.3,
2. Re-generated ODFDOM calling: mvn clean install -Ppedantic -Dcodegen
3. Tested the Taglets for attributes/elements/data-types and fixed the following anomalies:

## Unicode Local Change 
When I adjusted our existing ODF templates for new documents in ODFDOM from ODF 1.2 to ODF 1.3:
https://github.com/tdf/odftoolkit/tree/master/odfdom/src/main/resources
by using LibreOffice 7.4.3 (configured to use strict ODF 1.3) to load/save the templates.
I had to adjust the version of the ODI (Image) and ODG (Graphic) file manually.
(All validated as strict ODF 1.3 with https://odfvalidator.org/ )
Suddently, the there are by default Unicode Locales for the Unicode Groups and the three assertions return values:
https://github.com/tdf/odftoolkit/blob/master/odfdom/src/test/java/org/odftoolkit/odfdom/doc/DocumentTest.java#L437 The test was adopted.

## &lt;manifest:key-derivation&gt;
The element <manifest:key-derivation> has the choice of having either a single mandatory attribute value or multiple in ODF 1.3:

* ODF 1.3: https://docs.oasis-open.org/office/OpenDocument/v1.3/os/schemas/OpenDocument-v1.3-manifest-schema-rng.html#key-derivation-attlist
* ODF 1.2: https://github.com/tdf/odftoolkit/blob/master/generator/schema2template/src/test/resources/test-input/odf/grammar/OpenDocument-v1.2-os-manifest-schema.rng#L165

The code generator should better create two constructors one with a single and the other with three parameter, but it only provides the single paramater constructor, which occurred first. OdfPackage was adjusted for now, but code generator might better be improved.

## SNAPSHOT Release

A SNAPSHOT released has been deployed to the Maven repository:
https://oss.sonatype.org/content/groups/public/org/odftoolkit/odfdom-java/0.12.0-SNAPSHOT/ 

It has been 10 years since we updated an ODF version. Is there something further to be adjusted (some parts - like the grammars - have been prior already added to the ODF Validator and Generator project).

***NOTE***: The integration of this PR will fix and close #26.

PR #373, opened 2025-05-13
Chore/remove boxing and unboxing

xzel23:
This is quite huge. I had my IDE do this one for me, ran the tests and checked the diff. Looks good to me.


svanteschubert (approved):
@xzel23 Thanks for the patch, I see the pattern and made myself a note in #200 to adjust the templates for the generated files (in general, those beyond the dom java package).

PR #376, opened 2025-05-14
Feature/359 support java.time

xzel23:
Add support for java.time


xzel23:
So here we go:
- added support for java.time classes
- added setDateTime(Calendar) that sets date and time
- replaced the internal use of the pre-Java 8 date and time API
- replaced SimpleDateFormatter with DateTimeFormatter where necessary
- Calendar instances are always returned with the default time zone of the system

What I think needs to be reviewed: 
- Is the XML produced by setDateTime() valid? The tests pass, but I don't know how well they cover this.

PR #377, opened 2025-05-14
Chore/refactorings

xzel23:
- fix many "raw use of generics" warnings
-   MSVExpressionVisitorChildren contained to methods with th exact same implementation, I kept both, but delegate from the non-static one to the static one
- removed the class AlphabeticalOrder in AlphabeticalOrderedRunner.java, using a Comparator.comparing(...) instead


svanteschubert (approved):
Thanks for the patch, Axel!

PR #378, opened 2025-05-14
- fixed some "raw use of generics" warnings

xzel23:
- TextSelection implements Comparable<TextSelection>, removed raw use of Comparable from child classes
- small refactorings


svanteschubert (approved):
Very much more readable! 
Thank you, Axel!

PR #379, opened 2025-05-15
Chore/use primitives use bulk operations fix raw generics

xzel23:
Just some cleanup across the board. Let me know if the PRs get too big... maybe I should attack the module-info first, that one might take a while but should then be a rather small PR :)

Some notes:

# return of Collections

I changed the returns from the concrete implementation to the collection interface type. Returning the concrete type is an anti-pattern that exposes implementation details to the user that is none of their business and will prove problematic in the future.  This may break useer code if they use something like `LinkedList<...> = getSomeList();`, but it is easy to fix and every IDE I know of has a quick fix for that.

Declaring a return type Map instead of HashMap also has the advantage of being able to use a different implementation without the user even noticing. Examples are EnumMap (a map for EnumValues that offers much faster access or one of the ConcurrentMap implementations when thread safety is needed.

# LinkedLists

I changed all instantiations or LinkedList to ArrayList - while possible in theory, I have not seen a single case where LinkedList was really beneficial in the last 20 years. LinkedLists use more memory, and are slower to traverse. The only benefit would be frequent additions or removals at arbitrary positions inside the list. I profiled the tests and saw that LinkedList.indexOf() was one of the methods the most time was spent in. Even if both List implementations are O(n) here, the constant factor for ArrayList is much lower.

# String.intern()

Using String.intern() generally is a bad idea: the string will be copied into the constant pool and persist even after garbage collections. It was used for QNames and might lead to memory leaks, for example when running in a server context. Should there be any negative side effects, I can create a PR that fixes that without using intern(), but I think it should not be necessary.

# String comparisons using != and ==

These are simply bugs. It only works (more or less) reliably when __both__ strings are literals or have been obtained using intern(). It may or may not work if one of the strings is the result of a concatenation or was created using a StringBuilder/StringBuffer. IT may even work in all your tests and then fail after hours in your production system.

# redundant Map lookups

I have changed some places to reduce unnecessary map lookups, for example by iterating directly over values() instead of keys() and just doing a get(key) afterwards, or over the entySet() if both key and value are needed. I'll look for more places later.

### DIFF
--- a/generator/schema2template/src/main/java/schema2template/grammar/OdfFamilyPropertiesPatternMatcher.java
+++ b/generator/schema2template/src/main/java/schema2template/grammar/OdfFamilyPropertiesPatternMatcher.java
@@ -214,11 +214,10 @@ class OdfFamilyPropertiesPatternMatcher {
   @Override
   public String toString() {
     Map<String, List<String>> results = getFamilyProperties();
-    Set<String> families = results.keySet();
     StringBuilder sb = new StringBuilder();
-    for (String family : families) {
-      sb.append("@style:family = '").append(family).append("' =");
-      List<String> propNames = results.get(family);
+    for (Map.Entry<String, List<String>> entry : results.entrySet()) {
+      sb.append("@style:family = '").append(entry.getKey()).append("' =");
+      List<String> propNames = entry.getValue();
       for (String propName : propNames) {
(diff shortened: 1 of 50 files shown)


svanteschubert (approved):
I like your patch, it's just like a fresh breeze! 
Thank you for your contribution, Axel!
Svante

PR #382, opened 2025-05-18
refactor file contents tests; ignore macOS .DS_Store files in tests

xzel23:
I have changed the code that is used to compare directories:

- Instead of returning boolean values, it directly uses Assertion. methods to let JUnit do the formatting. I also renamed the methods from `compareDirectories()` to `assertDirectoriesEqual()` and made sure the order of parameters matches the standard order of the assert methods (expected result first, actual result second).
- When a difference in a text file is found, the original and changed lines are included in the assertion message.
- .DS_Store files are ignored - these files are created automatically by macOS when opening a directory in Finder; this means once you open a directory when looking for the cause of a failure, the tests will fail because now the number of files doesn't match and since these files are hidden by default, finding out what goes wrong may cause some headache.

### DIFF
--- a/generator/schema2template/src/test/java/schema2template/grammar/DirectoryCompare.java
+++ b/generator/schema2template/src/test/java/schema2template/grammar/DirectoryCompare.java
@@ -16,238 +16,159 @@
  */
 package schema2template.grammar;
 
-import java.io.BufferedReader;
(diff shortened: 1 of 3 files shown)


svanteschubert (approved):
Thanks for the refactoring, Axel!

PR #383, opened 2025-05-18
Chore/code cleanup

xzel23:
Sorry, this looks huge. But it's just small refactorings that bring the code up-to-date with Java 11 and should make it more readable.

- I fixed some errors and deprecations in Javadoc
- fixed all "raw use of parametrized class" warnings
- replaced loops with extended for loops
- replaced manual list-of-strings to string conversion by using Collectors.joining
- replaced creating and manual filling maps by bulk operations (`new HashMap<>(Map.of(...))`instead of just `Map.of(...)`is necessary to create mutable maps).
- some other small refactorings


svanteschubert (approved):
Thanks for your patch, Axel! :-)

PR #385, opened 2025-05-21
Fix/suspicious assignments

xzel23:
@svanteschubert here are too small changes I found by running some analytics on the code. The first looks like a bug to me, can you verify that? I don't have a test case for this, but when you look at the surrounding code, it is quite obvious. Unit tests pass with and without this change, so it seems the case just isn't covered.

The second one also is obvious: the assignment has no effect and the current code is just confusing.


svanteschubert (approved):
Thanks for the patch, obviously a copy/paste issue of sorting out the alignment:

image
see https://docs.oasis-open.org/office/OpenDocument/v1.3/os/part3-schema/OpenDocument-v1.3-os-part3-schema.html#property-text_anchor-type

PR #391, opened 2025-05-24
fix PROP_XXX constants, apply some refactorings

xzel23:
- fix invalid bitwise operations: 1 << (32+i) is the same as 1<< i, the shifted value needs to be long for this to work
- encapsulate "type" - it was mutable and public
- assign fieldMap directly, make it a static final constant and change the name accordingly; add getFieldMap() method for access


svanteschubert (approved):
Nicely done, thanks for the patch, Axel!

PR #400, opened 2025-06-27
replace legacy OSSRH stuff with Central

mistmist:
Sonatype is retiring OSSRH and the namespace was migrated to Maven Central already.

Try to adapt docs and pom.xml with snippets copied from various docs and hope it will work...

[... just typed "mvn deploy" and the result is at https://central.sonatype.com/service/rest/repository/browse/maven-snapshots/org/odftoolkit/ ]

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -228,13 +228,12 @@
     </ciManagement>
     <distributionManagement>
         <snapshotRepository>
-            <id>ossrh</id>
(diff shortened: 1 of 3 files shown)


svanteschubert (approved):
Thank you, Michael for taking the time on this.
We might need another iteration when the next full release will not work ;-) But this looks quite good to me!
PS: We should keep in mind that we still need to generate the HTML docu from the mdtext but this is no blocker for this patch to me!

